PR #136, opened 2026-02-23
Harden benchmark module import without weakening tests

MattMatheus:
## Summary
Fixes the pre-existing benchmark collection/import fragility without bypassing test rigor.

## Problem
`benchmarks.py` used top-level imports for benchmark-only dependencies (`autoscraper`, `mechanicalsoup`, `pyquery`, etc.).
When those are absent, module import fails during tooling contexts (including pytest doctest collection from repo root).

## What changed
- Reworked `benchmarks.py` to lazily import benchmark-only dependencies via `_import_or_raise(...)`.
- Renamed benchmark functions from `test_*` to `bench_*` so pytest no longer treats benchmarks as unit tests.
- Kept explicit runtime failure with actionable message when someone executes benchmarks without required benchmark deps.
- Preserved `__main__` benchmark runner behavior.

## Why this is rigorous
- No global ignore flags.
- No silent skipping of actual test files.
- Benchmarks are treated as benchmarks, not test suite members.
- Missing optional benchmark deps now fail only when benchmark execution is requested.

## Validation
- `python -m pytest -q benchmarks.py --collect-only` -> `0 tests collected` (expected)
- `.venv/bin/python -c "import benchmarks"` -> import succeeds in project env

### DIFF
--- a/benchmarks.py
+++ b/benchmarks.py
@@ -1,22 +1,33 @@
 import functools
 import time
 import timeit
+from importlib import import_module
 from statistics import mean
-
-import requests
-from autoscraper import AutoScraper
-from bs4 import BeautifulSoup
-from lxml import etree, html
-from mechanicalsoup import StatefulBrowser
-from parsel import Selector
-from pyquery import PyQuery as pq
-from selectolax.parser import HTMLParser
 
 from scrapling import Selector as ScraplingSelector
 
 large_html = (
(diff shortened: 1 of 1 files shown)


D4Vinci:
It was intended not include the import in the benchmarks, and also this PR is against the `main` branch instead of the `dev` branch, which means I can't accept it as stated in the contribution rules.

MattMatheus:
Sorry about that, I was testing something that wasn't quite ready for production use.  Won't happen again.

PR #137, opened 2026-02-24
fix: make MCP get schemas validator-safe

robin-ede:
## Summary
- Use MCP-friendly input types for `get` and `bulk_get` by changing `cookies`, `proxy_auth`, and `auth` to dictionary-based schemas.
- Add `_NormalizeCredentials` to convert auth dictionaries into tuple credentials expected by the fetchers.
- Update parameter docs to reflect the new dictionary format for basic auth fields.

## Why
Some strict MCP clients reject tuple/list-based JSON schema variants generated for these fields and fail tool registration with errors like `array schema missing items`. This change keeps runtime behavior while producing schemas that validate reliably across clients.

## Validation
- `python3 -m py_compile scrapling/core/ai.py`
- OpenCode `1.2.10` compatibility check:
  - Before this patch: MCP tool registration failed with `Invalid schema for function 'scrapling_get' ... array schema missing items`.
  - After this patch: Scrapling MCP tools register successfully.


D4Vinci:
Looks good to me, thanks mate!

Issue #167, opened 2026-03-06
Selector.get_all_text() doesn't get all text


mhillebrand:
### Have you searched if there an existing issue for this?

- [x] I have searched the existing issues

### Python version (python --version)

Python 3.13.12

### Scrapling version (scrapling.__version__)

0.4.1

### What's your operating system?

Linux Mint 22.2

### Are you using a separate virtual environment?

Yes

### Expected behavior

Expected all text to be retrieved:
```
string1
string2
string3
string4
string5
```

### Actual behavior

Not all text is retrieved:
```
string1
string2
string4
```

### Steps To Reproduce

```python
from scrapling.parser import Selector

html = """
<html>
<body>
    <main>
        string1
        <b>string2</b>
        string3
        <div>
            <span>string4</span>
        </div>
        string5
    </main>
</body>
</html>
"""

page = Selector(content=html)
node = page.css('main')[0]
print(node.get_all_text('\n', strip=True))
```

PR #168, opened 2026-03-06
fix: Selector.get_all_text() doesn't get all text #167

mhillebrand:
- Selector.get_all_text() no longer ignores tail text nodes
- Added new unit test

### Type of change:

- [ ] Dependency upgrade
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New integration (thank you!)
- [ ] New feature (which adds functionality to an existing integration)
- [ ] Deprecation (breaking change to happen in the future)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Add or change doctests? -- Note: Please avoid changing both code and tests in a single pull request.
- [ ] Documentation change?

### Additional information

- This PR fixes or closes an issue: fixes #167

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/D4Vinci/Scrapling/blob/main/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have doc-strings.

PR #169, opened 2026-03-06
fix: include tail text nodes in Selector.get_all_text

stablegenius49:
## Summary\n- include text node tails when collecting text so mixed content isn't dropped\n- preserve document order by iterating XPath text nodes\n- keep tail text that follows ignored tags while still excluding ignored tag content\n- add regression tests for nested-tag tails and ignored-tag tails\n\n## Validation\n- source .venv311/bin/activate && python -m pytest tests/parser/test_parser_advanced.py -k "get_all_text_includes_tail_nodes or get_all_text_keeps_ignored_tag_tail_text or text_operations_edge_cases"\n- source .venv311/bin/activate && python -m pytest tests/parser/test_general.py\n\nFixes #167

### DIFF
--- a/scrapling/parser.py
+++ b/scrapling/parser.py
@@ -304,13 +304,28 @@ class Selector(SelectorsGeneration):
                 ignored_elements.update(cast(list, _find_all_elements(element)))
 
         _all_strings = []
-        for node in self._root.iter():
-            if node not in ignored_elements:
-                text = node.text
-                if text and isinstance(text, str):
-                    processed_text = text.strip() if strip else text
-                    if not valid_values or processed_text.strip():
-                        _all_strings.append(processed_text)
+
+        def _append_if_valid(text: Any) -> None:
+            if text and isinstance(text, str):
+                processed_text = text.strip() if strip else text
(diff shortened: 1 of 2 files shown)


mhillebrand:
Did you not think my PR was sufficient? https://github.com/D4Vinci/Scrapling/pull/168

Avoiding recursion is preferable. I'm flattered that you used the unit test from my PR, tho!

D4Vinci:
A duplicate of #168 and #168 is the best approach because:
1. **Correctness**: The ancestor-walk approach for ignored tags is the only one that handles arbitrarily nested ignored elements correctly. This PR only checks parent/grandparent (breaks at depth 3+).
2. **Performance**: Using a pre-compiled XPath(".//text()") at the module level is consistent with the existing codebase patterns (__find_all_elements_, __find_all_elements_with_spaces_ are already pre-compiled the same way).
3. **Simplification**: It correctly removes the unnecessary _find_all_elements expansion from ignored_elements — since it walks ancestors, it only needs the tag elements themselves in the set, which is simpler and faster.

PR #175, opened 2026-03-07
fix: preserve tail text in Selector.get_all_text

### LINKED ISSUE #167: Selector.get_all_text() doesn't get all text

tsubasakong:
## Summary
- preserve tail text nodes when collecting recursive text content
- keep text order stable across nested children
- add regression coverage for interleaved child/tail text and ignored-tag tails

Fixes #167

## Validation
- source .venv/bin/activate && pytest tests/parser/test_parser_advanced.py -q
- source .venv/bin/activate && pytest tests/parser/test_general.py -q -k all_text
- git diff --check


D4Vinci:
A duplicate of #168, and this PR is against the contribution rules.

Also, PR #168 is the best approach because:
1. **Correctness**: The ancestor-walk approach for ignored tags is the only one that handles arbitrarily nested ignored elements correctly. This PR still recurses into ignored elements, potentially leaking their content.
2. **Performance**: Using a pre-compiled XPath(".//text()") at the module level is consistent with the existing codebase patterns (__find_all_elements_, __find_all_elements_with_spaces_ are already pre-compiled the same way).
3. **Simplification**: It correctly removes the unnecessary _find_all_elements expansion from ignored_elements — since it walks ancestors, it only needs the tag elements themselves in the set, which is simpler and faster.

PR #187, opened 2026-03-11
docs: Add Korean (한국어) README translation

greatsk55:
Added Korean translation of README at `docs/README_KR.md` and updated the language links in the main README.

Follows the same structure and format as existing translations (JP, CN, etc.).


D4Vinci:
Thanks!

PR #192, opened 2026-03-15
test(ai): add _normalize_credentials edge case coverage

Bortlesboat:
## What

Extends `tests/ai/test_ai_mcp.py` with a `TestNormalizeCredentials` class covering edge cases for the `_normalize_credentials` helper that were previously untested:

- `None` input → returns `None`
- Empty dict `{}` → returns `None` (falsy check)
- Valid dict with both keys → returns correct `(username, password)` tuple
- Dict missing `password` key → raises `ValueError`
- Dict missing `username` key → raises `ValueError`

## Why

The helper is used at every fetcher call site to validate proxy auth and request auth. The existing suite only exercises the happy path via integration tests — no unit tests verified the error branches or the empty-dict falsy path.

## Test

```bash
pytest tests/ai/test_ai_mcp.py::TestNormalizeCredentials -v
```

All 5 new tests pass. No existing tests affected.


D4Vinci:
Thanks!

PR #196, opened 2026-03-15
fix: replace bare raise with return False in _restore_from_checkpoint

haosenwang1018:
## What

Replace `raise` (bare raise with no active exception) with `return False` in `CrawlerEngine._restore_from_checkpoint()`.

## Why

When `_checkpoint_system_enabled` is `False`, the current code executes a bare `raise` statement. Since there is no active exception in this context, Python raises `RuntimeError: No active exception to re-raise`.

The method's docstring states: "Returns True if successfully restored, False otherwise." Returning `False` matches this contract and is consistent with how the caller in `crawl()` uses the return value.

The caller currently guards with `if self._checkpoint_system_enabled`, so this path isn't hit in normal usage — but the method should be self-consistent for any future callers.

## Changes

- `scrapling/spiders/engine.py`: 1 line changed (`raise` → `return False`)

## Type of change


D4Vinci:
Thanks!
Once thew workflows pass, we are ready to merge

PR #197, opened 2026-03-15
fix: add max retry limit to _get_page_content to prevent infinite loop

haosenwang1018:
## What

Add a bounded retry limit (default 10 attempts = ~5s) to `_get_page_content` and `_get_async_page_content` in `ResponseFactory`.

## Why

Both methods use `while True` to retry `page.content()` on `PlaywrightError`. This is a workaround for a known Windows issue ([playwright#16108](https://github.com/microsoft/playwright/issues/16108)) where `page.content()` can transiently fail.

However, if the page is in a permanently broken state (crashed tab, closed context, destroyed frame), the loop retries forever and hangs the entire process with no way to recover.

Replacing with a bounded `for` loop preserves the retry behavior for transient errors while preventing indefinite hangs.

## Changes

- `scrapling/engines/toolbelt/convertor.py`: Replace `while True` with `for _ in range(max_retries)` in both sync and async versions

## Type of change

### DIFF
--- a/scrapling/spiders/session.py
+++ b/scrapling/spiders/session.py
@@ -112,10 +112,12 @@ class SessionManager:
                 client = session._client
 
                 if isinstance(client, _ASyncSessionLogic):
+                    kwargs = request._session_kwargs.copy()
+                    method = cast(SUPPORTED_HTTP_METHODS, kwargs.pop("method", "GET"))
                     response = await client._make_request(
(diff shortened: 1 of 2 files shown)


D4Vinci:
That's a good idea here for handling this edge case, but it's not good practice not to tell the user what happened. I think it will be better to raise an error if the maximum number of retries is reached. I will add it. Thanks!

PR #200, opened 2026-03-16
test: add edge case tests for filter, iterancestors, and find_similar

awanawana:
## Summary

This PR adds three new test files covering edge cases that were missing from the existing test suite (current coverage ~90–92%).

#### `tests/parser/test_selectors_filter.py`
Covers `Selectors.filter()` edge cases:
- Chained `.filter().filter()` calls
- Empty result returns `Selectors` (not `None` or exception)
- All-pass predicate returns all elements
- Calling `filter()` on an already-empty `Selectors`

#### `tests/parser/test_ancestor_navigation.py`
Covers `Selector.iterancestors()` and `Selector.find_ancestor()`:
- Ancestors are yielded bottom-up (immediate parent first)
- All ancestors up to `<html>` are included
- `find_ancestor()` returns the closest match
- `find_ancestor()` returns `None` when no ancestor matches
- Both methods are safe on text nodes (no exception)
- `find_ancestor()` on root element returns `None` gracefully

#### `tests/parser/test_find_similar_advanced.py`
Covers `Selector.find_similar()` with non-default parameters:
- Higher `similarity_threshold` returns fewer or equal results
- `match_text=True` factors in text content
- `ignore_attributes` combinations affect matching
- Calling on a text node returns empty `Selectors` without raising

### Test results
All 16 tests pass locally:
```
16 passed in 0.42s
```


D4Vinci:
LGTM, thanks!

PR #201, opened 2026-03-16
fix: preserve HTTP method across retries in spider session

karesansui-u:
`SessionManager.fetch()` pops `method` from `_session_kwargs`,
which mutates the original request dict. When the engine retries
a blocked request via `request.copy()`, the copy no longer has
`method`, so it defaults to GET.

Steps to reproduce:
1. Yield `Request(url, method="POST", data=...)`
2. Target returns a response that triggers `is_blocked()`
3. Engine retries via `request.copy()` → second fetch uses GET

Fix: copy the kwargs dict before popping, so the original
request stays intact.


D4Vinci:
Thanks, nice catch!

PR #205, opened 2026-03-19
fix: two bugs in parser.py

yetval:
Spotted two bugs in `parser.py`:

**Bug 1** — `xpath()` `else` branch calls `__handle_elements(elements)` when `elements` is never assigned, causing an `UnboundLocalError`. Fixed by passing `[]` instead.

**Bug 2** — `__are_alike` increments `checks` by `len(candidate_attributes)` but iterates over `original_attributes`, giving the wrong denominator in similarity scoring. Fixed by using `len(original_attributes)`.

### DIFF
--- a/scrapling/parser.py
+++ b/scrapling/parser.py
@@ -987,7 +987,7 @@ class Selector(SelectorsGeneration):
                 SequenceMatcher(None, v, candidate_attributes.get(k, "")).ratio()
                 for k, v in original_attributes.items()
             )
-            checks += len(candidate_attributes)
+            checks += len(original_attributes)
         else:
             if not candidate_attributes:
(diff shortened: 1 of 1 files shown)


D4Vinci:
Hey, thanks for the PR, but I can't accept it for multiple reasons:
1. It's against the contribution rules (made against the `main` branch).
2. Both bugs are invalid:
   - The walrus operator on line 656: `if elements := self._root.xpath(selector, **kwargs):` binds elements even when the result is falsy (empty list). So by the time we reach line 684 in the else branch, elements is already []. There's no UnboundLocalError.
   - Using `len(candidate_attributes)` is actually intentional, as it penalizes candidates that have extra attributes the original doesn't have or have fewer attributes. Making them less "alike".

PR #225, opened 2026-04-03
fix: save checkpoint before cancel_scope.cancel() on force-stop to prevent data loss

voidborne-d:
## Problem

On force-stop (second Ctrl+C), `CrawlerEngine.crawl()` calls `tg.cancel_scope.cancel()` **before** `_save_checkpoint()`. In anyio, cancelling a scope causes all subsequent `await` checkpoints within that scope to raise `Cancelled`. This creates a three-step data loss sequence:

1. `_save_checkpoint()` uses `anyio.open_file()` and `temp_path.rename()` — both are await checkpoints that get cancelled immediately after `cancel_scope.cancel()`
2. `self.paused` is never set to `True` (the assignment follows the aborted save)
3. The `finally` block sees `not self.paused` and calls `checkpoint_manager.cleanup()`, which **deletes** the previous checkpoint file

**Result:** A user who ran a long crawl, pressed Ctrl+C twice to force-stop, loses their entire checkpoint irrecoverably. The old checkpoint (from periodic saves or a previous graceful pause) is deleted, and the new one was never written.

### Code before fix:

```python
if self._force_stop:
    log.warning(f"Force stopping with {self._active_tasks} active tasks")
    tg.cancel_scope.cancel()       # ← All subsequent awaits raise Cancelled

if self._checkpoint_system_enabled:
    await self._save_checkpoint()   # ← CANCELLED: never completes
    self.paused = True              # ← Never reached
    log.info("Spider paused, checkpoint saved")

# ... later in finally:
if not self.paused:                 # ← True (paused was never set)
    await self._checkpoint_manager.cleanup()  # ← DELETES old checkpoint
```

## Fix

Move `cancel_scope.cancel()` **after** the checkpoint save:

```python
if self._checkpoint_system_enabled:
    await self._save_checkpoint()   # ✅ Completes normally
    self.paused = True              # ✅ Set correctly

if self._force_stop:
    tg.cancel_scope.cancel()        # Now safe — checkpoint is saved
```

The checkpoint save completes normally, `self.paused` is set to `True`, and only then does the scope get cancelled to abort in-flight tasks. The `finally` block correctly sees `paused=True` and skips cleanup.

## Tests

Adds 6 regression tests (all passing alongside 63 existing engine tests):

| Test | What it verifies |
|------|------------------|
| `test_force_stop_saves_checkpoint_before_cancel` | Core regression: force-stop creates checkpoint file and sets paused=True |
| `test_graceful_pause_still_saves_checkpoint` | Single Ctrl+C (graceful) still works as before |
| `test_force_stop_checkpoint_is_loadable` | Checkpoint saved during force-stop is valid and deserializable |
| `test_normal_completion_cleans_up_checkpoint` | Normal exit still cleans up checkpoint (no regression) |
| `test_force_stop_without_checkpoint_system` | Force-stop without crawldir does not crash |
| `test_force_stop_preserves_existing_checkpoint` | Existing checkpoint from previous run is updated, not deleted |

## Files Changed

- `scrapling/spiders/engine.py` — reorder cancel_scope.cancel() after checkpoint save
- `tests/spiders/test_force_stop_checkpoint.py` — 6 new regression tests

### DIFF
--- a/scrapling/spiders/engine.py
+++ b/scrapling/spiders/engine.py
@@ -251,17 +251,21 @@ class CrawlerEngine:
                     while self._running:
                         if self._pause_requested:
                             if self._active_tasks == 0 or self._force_stop:
-                                if self._force_stop:
-                                    log.warning(f"Force stopping with {self._active_tasks} active tasks")
-                                    tg.cancel_scope.cancel()
-
-                                # Only save checkpoint if checkpoint system is enabled
+                                # Save checkpoint BEFORE cancelling the scope.
+                                # cancel_scope.cancel() makes all subsequent awaits
+                                # raise Cancelled, which would abort the checkpoint
+                                # write and leave self.paused=False — causing the
+                                # finally block to DELETE the previous checkpoint.
                                 if self._checkpoint_system_enabled:
                                     await self._save_checkpoint()
                                     self.paused = True
                                     log.info("Spider paused, checkpoint saved")
                                 else:
                                     log.info("Spider stopped gracefully")
+
+                                if self._force_stop:
+                                    log.warning(f"Force stopping with {self._active_tasks} active tasks")
+                                    tg.cancel_scope.cancel()
 
                                 self._running = False
                                 break
--- a/tests/spiders/test_force_stop_checkpoint.py
+++ b/tests/spiders/test_force_stop_checkpoint.py
@@ -1,0 +1,278 @@
+"""Tests for force-stop checkpoint preservation in CrawlerEngine.
+
+Regression tests for the bug where force-stop (second Ctrl+C) called
+cancel_scope.cancel() BEFORE saving the checkpoint, causing:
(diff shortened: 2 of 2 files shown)


D4Vinci:
Hey @voidborne-d 
Thanks for the PR, this bug is valid. I have checked the code, but this PR violates the contribution rules (it's against the `main` branch).
Please close it and make another PR against the `dev` branch so I can merge it.

D4Vinci:
Also, this PR is made against an older version of the `engine.py`file so we can't merge it without shifting the lines anyway

voidborne-d:
Thanks — I rebased the fix onto `dev`, updated it against the current `engine.py`, and opened the replacement PR here: #230. Closing this one per your request.

D4Vinci (commented):
Once you make this change, the PR should be ready for merge.

PR #230, opened 2026-04-05
fix: save checkpoint before cancel_scope.cancel() on force-stop to prevent data loss

voidborne-d:
## Problem

On force-stop (second Ctrl+C), `CrawlerEngine.crawl()` calls `tg.cancel_scope.cancel()` **before** `_save_checkpoint()`. In anyio, cancelling a scope causes all subsequent `await` checkpoints within that scope to raise `Cancelled`. This can abort checkpoint persistence and leave `self.paused=False`, which then lets the `finally` cleanup remove the previous checkpoint.

## Fix

- save checkpoint before cancelling the task group scope
- keep `self.paused = True` on checkpoint-backed pause/force-stop paths
- port the regression tests onto the current `dev` branch test scaffolding

## Tests

```bash
pytest -q tests/spiders/test_force_stop_checkpoint.py
```

(12 passed locally)

### DIFF
--- a/scrapling/spiders/engine.py
+++ b/scrapling/spiders/engine.py
@@ -330,17 +330,21 @@ class CrawlerEngine:
                     while self._running:
                         if self._pause_requested:
                             if self._active_tasks == 0 or self._force_stop:
-                                if self._force_stop:
-                                    log.warning(f"Force stopping with {self._active_tasks} active tasks")
-                                    tg.cancel_scope.cancel()
-
-                                # Only save checkpoint if checkpoint system is enabled
+                                # Save checkpoint BEFORE cancelling the scope.
+                                # cancel_scope.cancel() makes all subsequent awaits
+                                # raise Cancelled, which would abort the checkpoint
+                                # write and leave self.paused=False — causing the
+                                # finally block to DELETE the previous checkpoint.
                                 if self._checkpoint_system_enabled:
                                     await self._save_checkpoint()
                                     self.paused = True
                                     log.info("Spider paused, checkpoint saved")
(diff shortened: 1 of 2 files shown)


scrapling/spiders/engine.py
```diff
@@ -330,18 +330,22 @@ async def crawl(self) -> CrawlStats:
                     while self._running:
                         if self._pause_requested:
                             if self._active_tasks == 0 or self._force_stop:
-                                if self._force_stop:
-                                    log.warning(f"Force stopping with {self._active_tasks} active tasks")
-                                    tg.cancel_scope.cancel()
-
-                                # Only save checkpoint if checkpoint system is enabled
+                                # Save checkpoint BEFORE cancelling the scope.
```
D4Vinci, line 333: Please remove these comments or replace them with a one-liner.

The commit message and PR description already explain the "why" in detail. The code itself is clear enough, with the reorder. Anyone reading it can see that the checkpoint save happens before the cancel. 

A single line like `# Save checkpoint before canceling to avoid data loss` would be plenty, if anything at all.

D4Vinci (changes requested):
Thanks @voidborne-d LGTM.
Once you make this change, the PR will be merged.

scrapling/spiders/engine.py
D4Vinci: NVM, I will remove it myself since I'm already pushing another feature and don't want it to affect this PR.

D4Vinci (approved):

D4Vinci:
Thanks for spotting this one and the help!

PR #238, opened 2026-04-09
feat: add pre_nav_listeners parameter to browser fetchers (#237)

### LINKED ISSUE #237: ability to specify custom page listeners
### Have you searched if there an existing feature request for this?

- [x] I have searched the existing requests

### Feature description

Hi,

first off, thanks for the awesome tool, really well designed.
In other tools like Crawlee, you can set preNavigationHooks which allows you to set custom Playwright event listener and do whatever logic you need within them. they are registered before `page.goto()` is called. 
this is not limited to XHR/Fetch requests.. sometimes we need to set custom logic on 'websocket' like `page.on("websocket", async (ws) => {...do stuff...}`. 
this currently can't be achieved by "page_action" since it runs AFTER the navigation to the page. 
i suggest adding something like in _controllers.py and _stealth.py
```
# After existing response handler registration, before page.goto():
 if params.pre_nav_listeners:
     for event_name, handler in params.pre_nav_listeners.items():
         page.on(event_name, handler)
```
to be used like: `yield response.follow(href, pre_nav_listeners={"websocket": lambda ws: self.on_websocket(ws, response.url)}` or a similar manner. this makes it fine grained to specific Request() object unlike crawlee preNavigationHook that is applied to full crawler. 

However, a couple of points I am not entirely sure of:
1- Scrapling already registers `page.on("response", ...)` internally, so what if the user registered the same handler ? which should run first? 
2- what if the listener is IO heavy (like dumping specific items from the listener on DB in realtime), how can that be optimal and not slow the whole crawler. should it use a different thread? 
curious of what you think. 

Thanks!

ousamabenyounes:
## Proposed change

Adds a `pre_nav_listeners` parameter to all four browser fetcher `fetch()` methods (`DynamicSession`, `AsyncDynamicSession`, `StealthySession`, `AsyncStealthySession`).

Handlers registered via `pre_nav_listeners` are attached to the Playwright page **before** `page.goto()` is called, making it possible to intercept events like `"websocket"` that must be set up prior to the first navigation — something that was previously impossible with `page_action` (which runs post-navigation).

### Type of change:

- [x] New feature (which adds functionality to an existing integration)

### Additional information

- This PR fixes or closes an issue: fixes #237
- This PR is related to an issue: #237
- Link to documentation pull request: N/A

### Checklist:

## Files changed

| File | Change |
|------|--------|
| `scrapling/engines/_browsers/_types.py` | Added `pre_nav_listeners: Optional[Dict[str, Callable]]` to `PlaywrightSession` and `PlaywrightFetchParams` TypedDicts |
| `scrapling/engines/_browsers/_validators.py` | Added field to `PlaywrightConfig` struct and `_fetch_params` dataclass |
| `scrapling/engines/_browsers/_controllers.py` | Register listeners before `page.goto()` in `DynamicSession.fetch()` and `AsyncDynamicSession.fetch()` |
| `scrapling/engines/_browsers/_stealth.py` | Same in `StealthySession.fetch()` and `AsyncStealthySession.fetch()` |
| `tests/fetchers/test_validator.py` | 6 new tests verifying default value, acceptance, and registration order before goto() |

Generated by Claude Code
Vibe coded by ousamabenyounes


D4Vinci:
Hey, thanks for the PR! The feature idea is solid (and directly addresses #237), but I'm going to close this in favor of a different design.

The `pre_nav_listeners` dict approach is too limiting:

- No duplicate events: Dict keys are unique, so you can't register multiple handlers for the same event (e.g., two "request" listeners).
- Only `page.on()`: You can't do other pre-navigation setup like `page.route()`, or anything else that needs to happen before `goto()`.
- Missing `setdefault`: `_fetch_params` now requires `pre_nav_listeners` but `validate_fetch` never sets a default for it, so calling `session.fetch()` without passing `pre_nav_listeners` will crash at runtime.

I'll implement this as a `pre_navigation_action` callback instead. So it's the same pattern as `page_action`, but runs before navigation. This gives users full control:

```python
def setup(page):
    page.on("websocket", ws_handler)
    page.on("websocket", another_handler)  # multiple handlers, same event
    page.on("request", req_handler)
    page.route("**/*", route_handler)  # not possible with dict approach

session.fetch(url, pre_navigation_action=setup)
```
Simpler API, strictly more powerful, and consistent with the existing `page_action` pattern. Thanks for the effort!

PR #240, opened 2026-04-10
fix: use Union[int, float] instead of Annotated[int, float] for Seconds type

kuishou68:
## Problem

In `scrapling/engines/_browsers/_validators.py`, the `Seconds` type alias is incorrectly defined as:

```python
Seconds = Annotated[int, float, Meta(ge=0)]
```

Per [PEP 593](https://peps.python.org/pep-0593/) / `typing.Annotated`, **only the first argument is the type**; subsequent arguments are metadata annotations. This means `float` is silently treated as a metadata value, **not** a type union. The actual validated type is `int` only.

As a result, passing `wait=0.5` or `timeout=1.5` silently truncates the float value to int, leading to unexpected behaviour with no error raised.

## Fix

Use `Union[int, float]` as the first (type) argument:

```python
Seconds = Annotated[Union[int, float], Meta(ge=0)]
```

Also adds `Union` to the `from typing import` statement for backward compatibility (Python < 3.10).

## Changes

- `scrapling/engines/_browsers/_validators.py`
  - Add `Union` to `from typing import Annotated, Union`
  - Change `Seconds = Annotated[int, float, Meta(ge=0)]` → `Seconds = Annotated[Union[int, float], Meta(ge=0)]`

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>

### DIFF
--- a/scrapling/engines/_browsers/_validators.py
+++ b/scrapling/engines/_browsers/_validators.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Annotated
+from typing import Annotated, Union
 from functools import lru_cache
 from urllib.parse import urlparse
 from dataclasses import dataclass, fields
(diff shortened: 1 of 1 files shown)


D4Vinci:
Hey, good catch! The bug is real -- `Annotated[int, float, Meta(ge=0)]` does treat `float` as metadata rather than a type, so passing `wait=1.5` fails.

However, the proposed fix (`Union[int, float]`) doesn't actually work with msgspec -- `Meta(ge=0)` can't be applied to a Union type and throws `Can only set 'ge' on a numeric type.` The correct fix is simply `Annotated[float, Meta(ge=0)]`, since `int` is a subtype of float in Python, so msgspec accepts both.

I'll fix this on `dev` and credit you as co-author. Also, for future reference, PRs should target the `dev` branch, not the `main` (per CONTRIBUTING.md). Thanks for spotting this!

PR #241, opened 2026-04-11
fix: duplicate ID segments in full-path selector generation

sjhddh:
## Summary
Full-path CSS and XPath selectors produce duplicate ID segments for elements with `id` attributes.

**Before:**
```
body > #main > #main > #target > #target
//body/[@id='main']/[@id='main']/[@id='target']/[@id='target']
```

**After:**
```
body > #main > #target
//body/[@id='main']/[@id='target']
```

## Root cause
In `_general_selection()` (`scrapling/core/mixins.py`), the `selectorPath.append(part)` at line 50 runs unconditionally after both branches. When an element has an `id`, the selector is appended in the `if` branch (line 30), then again at line 50 — producing duplicates.

## Fix
One-line change: indent `selectorPath.append(part)` into the `else` branch so it only fires for elements without an `id`.

## Test plan
- [x] 91/92 tests pass (1 pre-existing async test failure unrelated to this change)
- [x] 2 new regression tests added:
  - `test_full_path_selector_no_duplicate_ids` — verifies no duplicate segments for id elements
  - `test_full_path_selector_mixed_id_and_no_id` — verifies mixed id/non-id paths


sjhddh:
Addressed both review comments in 273c8c2:

**XPath syntax fix:** Full-path XPath mode now emits `*[@id="..."]` instead of bare `[@id="..."]`, producing valid steps like `//body/*[@id="main"]/*[@id="target"]`. Short-path mode unchanged.

**Test assertion:** Added XPath evaluation assertion to verify the generated selector actually selects the correct element.

D4Vinci:
Hey, nice find! The bug is real, and the fix is clean. Two things before merging:

1. Target branch: This PR targets `main`, but should target `dev` per `CONTRIBUTING.md`. Could you retarget it?
2. The rest looks good -- minimal diff, correct fix, and the XPath `*[@id='x']` improvement for full-path mode is a nice catch too.

sjhddh:
Retargeted to `dev` as requested. Thanks for the review!

D4Vinci:
Thanks for spotting this and your fix!

PR #245, opened 2026-04-16
fix: prevent FetcherSession state corruption and lazy session close crash

yetval:
## Summary

- **Lazy session close crash**: `SessionManager.close()` called `__aexit__` on every session including lazy ones never opened. `FetcherSession.__aexit__` raises `RuntimeError: Cannot exit invalid session` when `_client is None`. Fix: skip sessions where `sid in _lazy_sessions and not _is_alive`.

- **State corruption on enter failure**: `_is_alive = True` and `_client` were set before `__enter__`/`__aenter__` completed. If inner enter raised, object was left in broken half-initialized state. Fix: wrap inner enter in try/except, roll back `_client = None` on failure, set `_is_alive` only after success.

## Repro

```python
import anyio
from scrapling.spiders.session import SessionManager
from scrapling.fetchers import FetcherSession

async def main():
    sm = SessionManager()
    sm.add("default", FetcherSession(), lazy=True)  # never used
    await sm.start()
    await sm.close()  # RuntimeError before fix, clean exit after

anyio.run(main)
```

## Files changed

- `scrapling/engines/static.py` — `FetcherSession.__enter__` and `__aenter__`
- `scrapling/spiders/session.py` — `SessionManager.close`


D4Vinci (approved):
LGTM

D4Vinci:
Nice catch as always, thanks, mate!

Issue #244, opened 2026-04-14
Making screenshots in MCP mode


Gospodin:
### Have you searched if there an existing feature request for this?

- [x] I have searched the existing requests

### Feature description

It would be nice to add possibility to make screenshots in MCP mode

D4Vinci:
Thanks for the idea. The new feature is added now and will be published with the new update

PR #246, opened 2026-04-16
feat(mcp): add screenshot tool for capturing page screenshots

voidborne-d:
## Summary

Implements #244 — adds a `screenshot` MCP tool that captures web page screenshots and returns them as base64-encoded image data.

### New types
- `ScreenshotModel` — response model with `url`, `image_data` (base64), `image_type` (png/jpeg), and `full_page` fields
- `ScreenshotType` — literal type for image format selection

### New methods on `ScraplingMCPServer`
- `screenshot()` — takes a screenshot of a URL, supporting both ad-hoc and session-based usage
- `_get_session_any_type()` — helper that looks up a session by ID without enforcing dynamic/stealthy type (screenshot works with either)

### How it works

The core challenge is that Scrapling's `fetch()` manages the page lifecycle internally — the page is closed in `_page_generator`'s finally block before we can interact with it. The solution uses the existing `page_action` callback mechanism: an async closure captures the screenshot bytes from the live Playwright page during the fetch, before the page is cleaned up.

**Ad-hoc mode** (no `session_id`): Creates a temporary `AsyncDynamicSession`, navigates to the URL, captures via `page_action`, and tears down.

**Session mode** (`session_id` provided): Reuses an existing persistent session (either dynamic or stealthy), passing `page_action` through the fetch kwargs.

### Supported options
- `image_type`: `"png"` (default) or `"jpeg"`
- `full_page`: capture full scrollable page vs viewport only
- `quality`: JPEG quality (0-100)
- `wait` / `wait_selector` / `network_idle`: page readiness controls
- All standard session creation params for ad-hoc mode

### Tests
6 new test cases in `tests/ai/test_ai_mcp.py`:
1. Ad-hoc PNG screenshot with magic byte verification
2. Ad-hoc JPEG screenshot with quality parameter
3. Full-page screenshot
4. Screenshot with persistent dynamic session
5. Screenshot with persistent stealthy session
6. Error handling for nonexistent session ID

All existing tests continue to pass (the 2 pre-existing `test_get_tool`/`test_bulk_get_tool` failures are a curl_cffi compatibility issue with the new `"safe"` redirect value, unrelated to this PR).

### DIFF
--- a/scrapling/core/ai.py
+++ b/scrapling/core/ai.py
@@ -1,4 +1,5 @@
 from uuid import uuid4
+from base64 import b64encode
 from asyncio import gather
 from datetime import datetime, timezone
 from dataclasses import dataclass, field
@@ -63,6 +64,18 @@ class SessionClosedModel(BaseModel):
     message: str = Field(description="A confirmation message.")
 
 
+ScreenshotType = Literal["png", "jpeg"]
+
+
+class ScreenshotModel(BaseModel):
+    """Screenshot result with base64-encoded image data."""
+
+    url: str = Field(description="The URL of the page that was captured.")
+    image_data: str = Field(description="Base64-encoded image data.")
+    image_type: ScreenshotType = Field(description="The image format: 'png' or 'jpeg'.")
+    full_page: bool = Field(description="Whether the screenshot captures the full scrollable page.")
+
+
 @dataclass
 class _SessionEntry:
     session: Any  # AsyncDynamicSession | AsyncStealthySession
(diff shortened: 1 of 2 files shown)


D4Vinci:
Hey, thanks a lot for taking the time to put this together, and for the thorough PR description and tests. I'm going to close this one and take the feature over myself, but I want to explain why, so it's not just a drive-by rejection.

The main issue is the return shape. Returning the screenshot as a base64 string inside a Pydantic model with `structured_output=True` means the screenshot reaches the client as text in a JSON field, which the LLM can't actually *see*, only "read" as opaque base64. For a screenshot tool that defeats the purpose, and a single PNG can easily balloon into 100K+ tokens of noise in the response. The MCP SDK already has `mcp.server.fastmcp.Image` for exactly this case, which produces a proper `ImageContent` block that the client renders as a real image. So the tool needs to return `Image(data=..., format=...)` directly and drop `structured_output=True` for this registration.

There are also a few smaller things I want to restructure while I'm in there:

- `_get_session_any_type` duplicates most of `_get_session`, I'd rather collapse them into `_get_session(sid, expected_type=None)`.
- When `session_id` is passed together with ad-hoc options like `proxy` / `useragent` / `extra_headers`, those are silently dropped today. I'd prefer to raise a conflict rather than hide it.

Really appreciate the contribution, and please don't let this discourage you from opening more. Your write-up made it very easy to review.

PR #255, opened 2026-04-26
fix: hash request kwargs and headers correctly

yetval:
Summary
Request fingerprint collisions: `Request.update_fingerprint()` previously hashed only kwarg names and lowercased header values. Distinct requests could collapse to the same fingerprint when `fp_include_kwargs` or `fp_include_headers` were enabled, which can silently break scheduler deduplication, cache replay, and checkpoint restore. Fix: hash kwarg names together with their values and preserve header values as-is.

Repro
from scrapling.spiders.request import Request

r1 = Request("https://example.com", timeout=1)
r2 = Request("https://example.com", timeout=2)

assert r1.update_fingerprint(include_kwargs=True) != r2.update_fingerprint(include_kwargs=True)

Files changed
scrapling/spiders/request.py — fingerprint kwargs/header handling


samrusani:
I did a quick local check against this patch and the behavior matches the PR description:

```py
from scrapling.spiders.request import Request

assert Request("https://example.com", timeout=1).update_fingerprint(include_kwargs=True) != Request("https://example.com", timeout=2).update_fingerprint(include_kwargs=True)
assert Request("https://example.com", headers={"X-Test": "A"}).update_fingerprint(include_headers=True) != Request("https://example.com", headers={"X-Test": "a"}).update_fingerprint(include_headers=True)
```

`tests/spiders/test_request.py` also passed locally for me after applying the patch (`28 passed`).

One small suggestion: it would be worth adding these as regression tests in `tests/spiders/test_request.py`, since the contribution guide asks bug fixes to include code that reproduces the bug. That would lock in both the kwargs-value collision and the header-value case handling.

yetval:
Hey @samrusani, I have went ahead and added tests. Thanks for the suggestions!

scrapling/spiders/request.py
```diff
@@ -97,15 +97,19 @@ def update_fingerprint(
         }
 
         if include_kwargs:
-            kwargs = (key.lower() for key in self._session_kwargs.keys() if key.lower() not in ("data", "json"))
-            data["kwargs"] = "".join(set(_convert_to_bytes(key).hex() for key in kwargs))
+            filtered_kwargs = {
+                key.lower(): str(value)
```
D4Vinci: Using `str(value)` for kwarg values is fragile if someone passes a non-primitive.

D4Vinci (changes requested):

scrapling/spiders/request.py
yetval: Thanks! Just fixed this.

D4Vinci (approved):
LGTM

D4Vinci:
Good catch as always @yetval 
We will merge once the checks pass again.

PR #268, opened 2026-05-06
feat(spiders): add Spider.max_items hard cap on items_scraped

witnium:
## What this changes

Adds a `Spider.max_items: int | None` field that the engine enforces directly. When `stats.items_scraped >= max_items`, the engine drains the scheduler, stops dispatching new tasks, waits for in-flight callbacks to drain, and exits cleanly. Default is `None` (unlimited), preserving existing behaviour for everyone who doesn't set it.

This is the equivalent of Scrapy's `CLOSESPIDER_ITEMCOUNT` — a primitive most crawl frameworks have because user-callback caps don't work reliably under concurrency.

## Why

User-side caps in `parse()` (the obvious "stop yielding follows when N items reached" pattern) are racy with `concurrent_requests > 1`. With concurrency 4 and a seed page that yields M follows, all M follows can be enqueued before any single parse callback notices the cap.

Real-world reproduction: I had a spider hitting a WordPress homepage (78 outgoing links) with `concurrent_requests=4`, asking for `max_pages=10`. The local check produced **`items_scraped=40`** — 4× overshoot, and ~178s wall time on requests that were going to be discarded anyway.

After the patch, the same setup yields **`items_scraped=10`**, **`elapsed_seconds=5.0`** — the cap is exact and the crawl stops on time.

## How it works

* **`scrapling/spiders/spider.py`**: new optional class field `max_items: int | None = None`.
* **`scrapling/spiders/scheduler.py`**: new `Scheduler.drain()` helper. Drops every queued request without dequeuing them through the awaitable path. Returns the discarded count. Leaves the seen-fingerprints set intact so duplicates that slip through in-flight callbacks still get filtered.
* **`scrapling/spiders/engine.py`**:
  * Top of the `crawl()` loop: when `_max_items_reached()`, drain the scheduler once, log the drop count + in-flight count, then idle until `_active_tasks == 0` and exit.
  * In `_run_callbacks`: when the cap is hit, drop yielded `Request` follows (not enqueued) and additional `dict` items (counted as `items_dropped`). Caps overshoot from in-flight callbacks; without it, items can still slip in past the cap.
  * `_max_items_reached()` uses `getattr(spider, 'max_items', None)` so duck-typed test stubs that don't subclass `Spider` keep working.

In-flight overshoot: items can exceed the cap by up to `concurrent_requests - 1` because parses already dispatched at the moment the cap is hit will still complete their first item yield. Documented in the spider docstring; the `_run_callbacks` filter keeps it bounded to that.

## Tests

* `TestSchedulerDrain` (3 cases) — drain empties the queue, returns count, leaves the seen set intact, no-ops on empty.
* `TestRunCallbacks` (3 new cases):
  - drops yielded follows once the cap is reached
  - drops yielded items past the cap, counted as `items_dropped`
  - `max_items=None` preserves legacy behaviour
* `MockSpider` gains a `max_items` field so tests can opt in.

All 140 spider tests (test_engine + test_scheduler + test_spider) pass locally on Python 3.12.

## CI gates run locally

* ✅ `ruff check scrapling/` — all checks passed
* ✅ `ruff format --check scrapling/` — 47 files already formatted
* ✅ `vermin -t=3.10- --violations --eval-annotations --no-tips scrapling/` — minimum 3.10
* ✅ `bandit -r -c .bandit.yml scrapling/` — no issues
* ✅ `pytest tests/spiders/` — 140 passed

## Checklist

- [x] PR targets `dev` branch
- [x] Tests added for the new feature
- [x] Reproducer included (see "Why" above)
- [x] Backwards-compatible (`max_items=None` default = current behaviour)

### DIFF
--- a/scrapling/spiders/engine.py
+++ b/scrapling/spiders/engine.py
@@ -138,12 +138,24 @@ class CrawlerEngine:
         if not request.sid:
             request.sid = self.session_manager.default_session_id
 
+    def _max_items_reached(self) -> bool:
+        # getattr-defensive so a duck-typed spider stub (e.g. test fixtures
+        # that don't subclass `Spider`) works without setting this field.
+        cap = getattr(self.spider, "max_items", None)
+        return cap is not None and self.stats.items_scraped >= cap
+
     async def _run_callbacks(self, request: Request, response: Response) -> None:
         """Dispatch response to the request's callback and process yielded items/requests."""
         callback = request.callback if request.callback else self.spider.parse
         try:
             async for result in callback(response):
                 if isinstance(result, Request):
+                    # Stop accepting new follows once max_items is reached.
+                    # In-flight callbacks may keep yielding; just drop them
+                    # so the queue doesn't grow past the cap.
+                    if self._max_items_reached():
+                        log.debug(f"Dropped follow past max_items cap: {result.url}")
+                        continue
                     if self._is_domain_allowed(result):
                         self._normalize_request(result)
                         await self.scheduler.enqueue(result)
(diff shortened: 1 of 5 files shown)


D4Vinci:
I see this as a very rare use case that is not worth adding. Users who want this can do it better themselves.

PR #272, opened 2026-05-06
fix: support MCP user data dir env

### LINKED ISSUE #269: [Feature Request]: MCP server should support a config option for persistent cookie storage via `user_data_dir`
### Have you searched if there an existing feature request for this?

- [x] I have searched the existing requests

## Summary

The MCP server currently doesn't expose `user_data_dir`, so every session uses a temporary browser profile. Cookies and login state are lost when the session ends.

It would be very useful to have an environment variable (e.g. `SCRAPLING_USER_DATA_DIR`) that sets a persistent browser profile directory, so cookies survive across MCP sessions.

## Current Behavior

- `open_session` / `stealthy_fetch` / `fetch` — all create temporary profile dirs
- Underlying Python API (`StealthySession`, etc.) already supports `user_data_dir`, but MCP layer doesn't pass it through

## Proposed Solution

Add an optional env var like `SCRAPLING_USER_DATA_DIR=/path/to/profile` that, when set, is used as the `user_data_dir` for all MCP browser sessions. This way:

1. Login once → cookies persist on disk
2. Subsequent sessions automatically restore auth state
3. No need to re-inject cookies via the `cookies` param every time

## Workaround

Currently have to bypass MCP and use the Python API directly:

```python
from scrapling.fetchers import StealthySession

async with StealthySession(
    headless=True,
    user_data_dir="/path/to/profile",
) as session:
    page = await session.fetch("https://example.com/dashboard")

Bortlesboat:
## Summary
- read `SCRAPLING_USER_DATA_DIR` in the MCP server
- pass that profile path into dynamic and stealthy browser sessions created by `open_session`, `fetch`, and `stealthy_fetch`
- add unit coverage with fake browser sessions so the env plumbing is tested without launching Playwright

Fixes #269.

## Tests
- `py -3 -m pytest tests/ai/test_ai_mcp.py -q -k user_data_dir`
- `py -3 -m pytest tests/ai/test_ai_mcp.py -q`
- `py -3 -m compileall -q scrapling/core/ai.py tests/ai/test_ai_mcp.py`
- `git diff --check`

### DIFF
--- a/scrapling/core/ai.py
+++ b/scrapling/core/ai.py
@@ -1,3 +1,4 @@
+from os import environ
 from uuid import uuid4
 from asyncio import gather
 from datetime import datetime, timezone
@@ -33,6 +34,7 @@ from scrapling.core._types import (
 
 SessionType = Literal["dynamic", "stealthy"]
 ScreenshotType = Literal["png", "jpeg"]
+USER_DATA_DIR_ENV = "SCRAPLING_USER_DATA_DIR"
 
 
 class ResponseModel(BaseModel):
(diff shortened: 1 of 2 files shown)


D4Vinci:
Hi @Bortlesboat 
Sorry, this feature is not planned for addition. I will close this issue too

PR #273, opened 2026-05-06
fix(security): prevent shell injection via PR title in release workflow

aaronjmars:
## Summary

`.github/workflows/release-and-publish.yml` interpolates `github.event.pull_request.title` directly into shell `run:` blocks via `${{ ... }}`. A merged PR with a crafted title can execute arbitrary commands inside the release job, which holds `contents: write` and `id-token: write` and publishes to PyPI.

## Impact

The injection points are:

- L27: `run: echo "title=${{ github.event.pull_request.title }}" >> $GITHUB_OUTPUT`
- L39: `PR_TITLE="${{ steps.pr_title.outputs.title }}"`

A title such as `v1.0.0"; <attacker payload> #` is substituted into the shell line before bash sees it. Once the job is mid-run with the OIDC token minted and `GITHUB_TOKEN` available, an attacker can:

- tamper with the wheel/sdist after `python -m build` and before `pypa/gh-action-pypi-publish` uploads to PyPI (supply-chain compromise of a 46k★ package),
- exfiltrate `GITHUB_TOKEN` (write to repo) or the PyPI OIDC token,
- push commits to `main` directly.

The `if: github.event.pull_request.merged == true` gate doesn't block the injection — it only delays it until after maintainer merge, which is the same trust boundary already in play for any PR-merge workflow. A title that looks like a normal version (e.g. `v1.0.0` followed by something visually plausible in a busy review) is enough.

## Location

`.github/workflows/release-and-publish.yml:27` and `:39`

## Fix

- Pass `pull_request.title` through the step's `env:` block so it stays a shell variable (`$PR_TITLE`) and is never expanded by the YAML templater into the script body.
- Use the multiline `<<EOF` form of `$GITHUB_OUTPUT` so titles containing `=` or newlines round-trip safely.
- Tighten the version regex from `^v` (which accepts `vfoo`) to `^v[0-9]+(\.[0-9]+){1,2}([.-][a-zA-Z0-9.+-]+)?$`. Existing tags (`v0.4`, `v0.4.7`, `v0.3.14`) still match. The regex is no longer the security boundary — the env-var indirection is — but a tighter regex is defense in depth.

## Detected by

Aeon + semgrep (`yaml.github-actions.security.run-shell-injection.run-shell-injection`).
- Severity: high
- CWE-78 (OS Command Injection) / CWE-94 (Code Injection)

## Verification

- `actionlint` clean on the patched file.
- Existing tag set (`v0.4`, `v0.4.1` … `v0.4.7`, `v0.3.13`, `v0.3.14`) all match the new regex; injection-shaped titles (`v1.0.0"; curl evil`, `v$(echo)`, `vfoo`) all fail it.
- Output round-trip preserved: subsequent `${{ steps.pr_title.outputs.title }}` and `${{ steps.extract_version.outputs.version }}` references still resolve.

Happy to adjust the regex shape if the maintainer prefers stricter semver.

---
Filed by [Aeon](https://github.com/aaronjmars/aeon-aaron).

### DIFF
--- a/.github/workflows/release-and-publish.yml
+++ b/.github/workflows/release-and-publish.yml
@@ -22,10 +22,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Get PR title
-        id: pr_title
-        run: echo "title=${{ github.event.pull_request.title }}" >> $GITHUB_OUTPUT
-
       - name: Save PR body to file
         uses: actions/github-script@v8
         with:
@@ -35,13 +31,14 @@ jobs:
 
       - name: Extract version
         id: extract_version
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
-          PR_TITLE="${{ steps.pr_title.outputs.title }}"
-          if [[ $PR_TITLE =~ ^v ]]; then
-            echo "version=$PR_TITLE" >> $GITHUB_OUTPUT
+          if [[ "$PR_TITLE" =~ ^v[0-9]+(\.[0-9]+)*([.-]?[a-zA-Z][a-zA-Z0-9.+-]*)?$ ]]; then
+            printf 'version=%s\n' "$PR_TITLE" >> "$GITHUB_OUTPUT"
             echo "Valid version format found in PR title: $PR_TITLE"
           else
-            echo "Error: PR title '$PR_TITLE' must start with 'v' (e.g., 'v1.0.0') to create a release."
+            echo "Error: PR title must look like v1, v1.0, v1.0.0, v1.0.0-rc1, or v1.0.0rc1 to create a release."
             exit 1
           fi
 


yetval:
Fix looks good. One thing — targeted at `main`, should be `dev`. Mind rebasing?

Few small questions:
- `PR_TITLE_EOF` heredoc delimiter is literal; worth randomizing?
- `pr_title` step looks redundant now — could `extract_version` just env-pass the title directly?
- Regex rejects bare `v1` and `v1.0.0rc1` (no separator) — intentional?

aaronjmars:
Thanks for the careful review — addressed all of these in the latest push.

**Rebased onto `dev` and retargeted the PR base.**

**On the questions:**

1. **`PR_TITLE_EOF` heredoc delimiter** — good catch. The simpler fix was to remove the heredoc entirely: the `pr_title` step's only consumer was `extract_version`, so I collapsed them (see #2). The remaining `$GITHUB_OUTPUT` write is `printf 'version=%s\n' "$PR_TITLE"` *after* the regex has constrained `PR_TITLE` to `[a-zA-Z0-9.+-]` — no newlines or `=` possible, so a plain `key=value` line is safe and no delimiter is needed.

2. **`pr_title` step redundant** — agreed and removed. `extract_version` now env-passes `github.event.pull_request.title` directly. One step, one env-var indirection, one regex gate.

3. **Regex rejecting `v1` / `v1.0.0rc1`** — not intentional, just over-tight. Loosened to `^v[0-9]+(\.[0-9]+)*([.-]?[a-zA-Z][a-zA-Z0-9.+-]*)?$`, which now accepts `v1`, `v1.0`, `v1.0.0`, `v1.0.0-rc1`, `v1.0.0.rc1`, `v1.0.0rc1`, `v1.0.0a1`, `v2.0.0-beta.1`. All existing tags (`v0.4`, `v0.4.1`…`v0.4.7`, `v0.3.13`, `v0.3.14`) still match; injection-shaped titles (`v1.0.0"; curl evil`, `v$(echo)`, `vfoo`, `v1.0.0;rm`, `` v1.0.0`pwd` ``) still fail.

Net diff is now `+5 / -8`. `actionlint` clean.

D4Vinci:
It is unrealistic to have this issue, and it was reported to me before. The release workflow is only controlled by me, and it runs on GitHub CI only.

PR #281, opened 2026-05-14
Fix: cdp connections spawn isolated contexts instead of reusing existing profiles

Ewal11:
### What does this PR do?
This PR modifies the default behavior of `cdp_url` connections across all Fetcher variants (Sync and Async, Dynamic and Stealthy).

Currently, when passing a `cdp_url` to connect to an existing, headful Chrome browser, Scrapling unconditionally calls `browser.new_context()`. Because of how Playwright operates, this generates a completely isolated, incognito-like browsing context. As a result, the user's active logins, installed extensions, and session cookies from their main profile are not accessible to the Fetcher.

This PR checks if `browser.contexts` exists upon a successful CDP connection. If the default context exists, it assigns it to `self.context` instead of spawning a new one.

### Why is this needed?
The primary use case for connecting to an active browser via a debug port (`cdp_url`) is to leverage the existing, authenticated state of that browser (e.g., bypassing logins or utilizing active extensions). Spawning an isolated context defeats the purpose of connecting to a pre-configured user-data directory.

### Impact
* `cdp_url` connections will now inherit the browser's active tabs, cookies, and extensions.
* Proxy rotation and standard headless/headful launches are completely unaffected by this change.

### DIFF
--- a/scrapling/engines/_browsers/_controllers.py
+++ b/scrapling/engines/_browsers/_controllers.py
@@ -78,7 +78,10 @@ class DynamicSession(SyncSession, DynamicSessionMixin):
                 if self._config.cdp_url:  # pragma: no cover
                     self.browser = self.playwright.chromium.connect_over_cdp(endpoint_url=self._config.cdp_url)
                     if not self._config.proxy_rotator and self.browser:
-                        self.context = self.browser.new_context(**self._context_options)
+                        if self.browser.contexts:
+                            self.context = self.browser.contexts[0]
+                        else:
+                            self.context = self.browser.new_context(**self._context_options)
                 elif self._config.proxy_rotator:
                     self.browser = self.playwright.chromium.launch(**self._browser_options)
                 else:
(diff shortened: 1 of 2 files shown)


yetval:
Hey @Ewal11, this fix targets a real problem, but it creates more issues than it solves.

First, the PR is against `main`, but per `CONTRIBUTING.md` it should target `dev`

More importantly, reusing the user’s existing context breaks teardown: `close()` unconditionally calls `self.context.close()`, which now closes the user’s real Chrome tabs on every clean exit.

It also silently drops the configured `proxy` on the reuse path, so `cdp_url + proxy` can leak the real IP, which contradicts the PR description. The same applies to the other context options — `viewport`, `user_agent`, `extra_http_headers`, `color_scheme`, and `permissions` are all skipped when reusing a context, weakening Stealthy.

This needs an ownership flag, so teardown only closes contexts Scrapling created, and context options are only applied when Scrapling owns the context. Also, no tests were added, and all CDP paths are currently marked `# pragma: no cover`.

Not a maintainer, this is just to inform you!

D4Vinci:
Hi, thanks for the PR.

Since Chrome 136, `--remote-debugging-port` is refused on the default profile dir for security, so anyone using CDP already has to start Chrome with a dedicated `--user-data-dir` anyway. Regarding the "I want my logged-in profile" use case, that's exactly what the `user_data_dir` parameter is for.

Another big problem with your PR other than what @yetval mentioned is that blindly grabbing `browser.contexts[0]` skips `_context_options` entirely.

Issue #282, opened 2026-05-15
Unable to connect MCP server to Claude Desktop (docker approach)


sanjarcode:
### Have you searched if there an existing issue for this?

- [x] I have searched the existing issues

### Python version (python --version)

3.10.20

### Scrapling version (scrapling.__version__)

latest

### Dependencies version (pip3 freeze)

??

### What's your operating system?

MacOS Tahoe 26.04

### Are you using a separate virtual environment?

Yes

### Expected behavior

MCP should instlal

### Steps To Reproduce

- Follow https://scrapling.readthedocs.io/en/latest/ai/mcp-server.html, run `docker pull pyd4vinci/scrapling`
- Edit `claude_desktop_config.json` file accordingly
- Restart Claude Code
- You can see the error then

<img width="1152" height="525" alt="Image" src="https://github.com/user-attachments/assets/3c97ae15-012b-4bd0-874b-920b48976d92" />

The docker fails exits when started. And Claude Code shows an error:
<details><summary> Claude Desktop MCP error log (click to view)</summary>
```
2026-05-15T07:35:03.795Z [ScraplingServer] [info] Initializing server... { metadata: undefined }
2026-05-15T07:35:03.858Z [ScraplingServer] [info] Using MCP server command: /usr/local/bin/docker with args and path: {
  metadata: {
    args: [ 'run', '-i', '--rm', 'scrapling', 'mcp', [length]: 5 ],
    paths: [
      '/opt/local/bin',
      '/opt/local/sbin',
      '/Users/sanjar.afaq/.local/bin',
      '/Users/sanjar/.antigravity/antigravity/bin',
      '/opt/homebrew/bin',
      '/Users/sanjar.afaq/.cargo/bin',
      '/Users/sanjar.afaq/Library/pnpm',
      '/Users/sanjar.afaq/Library/Android/sdk/platform-tools',
      '/Users/sanjar.afaq/Library/Android/sdk/tools',
      '/usr/local/bin',
      '/usr/local/sbin',
      '~/bin',
      '/opt/homebrew/opt/libpq/bin',
      '/Users/sanjar.afaq/.rbenv/shims',
      '/Users/sanjar.afaq/.pyenv/shims',
      '/opt/homebrew/sbin',
      '/Users/sanjar.afaq/.local/share/zinit/polaris/bin',
      '/System/Cryptexes/App/usr/bin',
      '/usr/bin',
      '/bin',
      '/usr/sbin',
      '/sbin',
[...]

PR #283, opened 2026-05-15
docs: fix Docker image name in MCP example

yetval:
Docs at `docs/ai/mcp-server.md` instruct `docker pull pyd4vinci/scrapling`, but the Claude Desktop example config uses bare image name `scrapling`. Docker then tries the official library namespace and fails with `pull access denied for scrapling`.

Fix: use `pyd4vinci/scrapling` in the example so it matches the pulled image.

Verified locally — pulled `pyd4vinci/scrapling`, piped a JSON-RPC `initialize` into `docker run -i --rm pyd4vinci/scrapling mcp`, server returned `{"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2025-11-25",...,"serverInfo":{"name":"Scrapling","version":"1.27.1"}}}`.

Reported in #282.


D4Vinci:
Thanks mate!

PR #304, opened 2026-05-30
fix: apply session-level proxy when no per-request proxy is given (#295)

### LINKED ISSUE #295: [Bug]Session-level proxy silently ignored, leaks real IP
### Summary
`FetcherSession(proxy=...)` is silently ignored. Requests go direct and leak the real client IP. Passing `proxy=` per request on `.get()`/`.post()` works fine, only the session-level config is broken. Same bug on `AsyncFetcherClient`.

### Reproduce
```python
from scrapling.fetchers import FetcherSession
with FetcherSession(proxy='http://10.255.255.1:9999', timeout=8, retries=1) as s:
    r = s.get('http://httpbin.org/ip')
    print(r.status, r.body)
```
Expected: timeout, since `10.255.255.1` is unroutable.
Actual: `200 b'{"origin": "<my real public IP>"}'`.

### Root cause
In `scrapling/engines/static.py`, `_make_request` always forwards `proxy=proxy` (often `None`) into `_merge_request_args`, so the session default never gets used:
```python
"proxy": self._get_param(method_kwargs, "proxy", self._default_proxy),
```
The key is always present in `method_kwargs`, so the fallback branch is dead.

### Fix
Only pass `proxy` into `_merge_request_args` when it's not `None`. Same thing needs fixing in the async path. Probably worth checking `proxies` and `proxy_auth` while you're in there.

### Version
0.4.8, reproduces on `main` and `dev`.

ETM-Code:
## The bug (#295)

`FetcherSession(proxy=...)` is silently ignored: requests go direct and leak the real client IP. Passing `proxy=` per request works, only the session-level default is broken. Same on the async path.

```python
from scrapling.fetchers import FetcherSession
with FetcherSession(proxy="http://10.255.255.1:9999", timeout=8, retries=1) as s:
    r = s.get("http://httpbin.org/ip")
    print(r.status, r.body)   # expected: timeout. actual: 200 {"origin": "<real public IP>"}
```

## Root cause

`_make_request` does `static_proxy = kwargs.pop("proxy", None)` and then always re-passes `proxy=<value-or-None>` into `_merge_request_args`. Since `_get_param` treats a present-but-`None` key as an explicit value, the `self._default_proxy` fallback is dead. The underlying issue is that the code conflates "argument omitted" with "argument is `None`".

## Fix

Resolve the effective proxy in a single shared `_ConfigurationLogic._resolve_proxy` helper used by both the sync and async `_make_request`, with explicit precedence:

1. an explicit per-request `proxy` (including `None` to deliberately disable proxying)
2. the next proxy from the rotator, when one is configured
3. the session-level default `proxy`

A module-level `_UNSET` sentinel distinguishes an omitted `proxy` from an explicit `proxy=None`, removing the conflation that caused the leak. No new dependencies.

## Tests

Added sync + async coverage:
- omitted proxy inherits the session default (the regression itself)
- a per-request proxy overrides the session proxy
- explicit `proxy=None` disables proxying
- `proxy=None` skips the rotator
- the rotator yields a fresh proxy on each retry attempt

All assert on the `proxy` argument that actually reaches `curl_cffi`'s `session.request`. Reverting the fix fails the regression tests.

## Notes / scope

- One intentional behavior change: with a rotator configured, an explicit per-request `proxy=None` now means "no proxy for this request" instead of "use the rotator". An *omitted* proxy still uses the rotator.
- **Out of scope (follow-up):** this PR fixes the singular `proxy` only. The plural `proxies` mapping is resolved separately in `_merge_request_args` and already honors the session default, but the `proxy`/`proxies` interaction (e.g. an explicit per-request `proxy` not suppressing a session-level `proxies`) is a pre-existing design question I left untouched here. Happy to follow up if you'd like it handled.
- Targets `dev` per CONTRIBUTING.md. Ruff, ruff-format, vermin (`-t=3.10-`), and bandit pass on the changed code; the fetcher test suites pass locally.

Closes #295

### DIFF
--- a/scrapling/engines/static.py
+++ b/scrapling/engines/static.py
@@ -30,6 +30,8 @@ from ._browsers._types import RequestsSession, GetRequestParams, DataRequestPara
 from .toolbelt.fingerprints import generate_headers, __default_useragent__
 
 _NO_SESSION: Any = object()
+# Sentinel for "argument not provided", so an explicit ``None`` can be told apart from omission.
+_UNSET: Any = object()
 
 
 def _select_random_browser(impersonate: ImpersonateType) -> Optional[BrowserTypeLiteral]:
@@ -97,6 +99,25 @@ class _ConfigurationLogic(ABC):
     def _get_param(kwargs: Dict, key: str, default: Any) -> Any:
         """Get parameter from kwargs if present, otherwise return default."""
         return kwargs[key] if key in kwargs else default
+
+    def _resolve_proxy(self, request_proxy: Any = _UNSET) -> Any:
+        """Resolve the effective singular ``proxy`` for a single request.
+
+        Precedence, highest first:
+        1. An explicit per-request ``proxy`` (including ``None`` to drop the session ``proxy``).
+        2. The next proxy from the rotator, when one is configured.
+        3. The session-level default ``proxy``.
+
+        ``request_proxy`` stays ``_UNSET`` when the caller omitted it, so an explicit
+        ``proxy=None`` is not mistaken for "unset" and the session default is applied only
+        when no per-request proxy was given (the conflation behind #295). The plural
+        ``proxies`` mapping is resolved separately in :meth:`_merge_request_args`.
+        """
+        if request_proxy is not _UNSET:
+            return request_proxy
+        if self._proxy_rotator:
+            return self._proxy_rotator.get_proxy()
+        return self._default_proxy
 
     def _merge_request_args(self, **method_kwargs) -> Dict[str, Any]:
         """Merge request-specific arguments with default session arguments."""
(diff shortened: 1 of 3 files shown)


141319co:
Validated this patch locally on macOS with Python 3.12.13.

Checks run successfully:
- `python -m pytest tests/fetchers/sync/test_requests_session.py tests/fetchers/async/test_requests_session.py -q` -> 14 passed
- `python -m pytest tests/fetchers/test_proxy_rotation.py -q` -> 39 passed
- `ruff check scrapling/engines/static.py tests/fetchers/sync/test_requests_session.py tests/fetchers/async/test_requests_session.py` -> passed
- `ruff format --check scrapling/engines/static.py tests/fetchers/sync/test_requests_session.py tests/fetchers/async/test_requests_session.py` -> passed
- `vermin -t=3.10- --violations --eval-annotations --no-tips ...` -> passed
- `bandit -r -c .bandit.yml scrapling/engines/static.py tests/fetchers/sync/test_requests_session.py tests/fetchers/async/test_requests_session.py` -> no issues

The `_UNSET` sentinel approach looks like the right fix for distinguishing omitted `proxy` from explicit `proxy=None`, and the regression coverage checks sync, async, per-request override, explicit disable, and retry rotation behavior.

D4Vinci:
Hey, thanks for the PR and the solid test coverage! While reviewing it, we realized the root cause could be fixed with the existing fallback idiom in two lines, without a sentinel or new semantics for explicit `proxy=None`, so we went with that.

The fix ships in 0.4.9, and thanks again!

PR #319, opened 2026-06-01
docs: fix re_first docstring to match implementation

viraatdas:
The `re_first` docstring said it calls `.re_first()` on each element, but the implementation runs the given regex against each element and returns the first match (or the default value). Updated the docstring to describe the actual behavior.

Documentation only, no logic change. Verified that `tests/parser/` passes and `ruff check scrapling/core/custom_types.py` is clean.

Prepared with AI assistance.


D4Vinci:
Hey, the docstring is correct!

PR #323, opened 2026-06-02
fix: parse quoted charset values in content-type headers

Bortlesboat:
## Problem

`ResponseFactory.__extract_browser_encoding` extracts the charset from a `content-type` header with:

```python
__CHARSET_RE__ = re_compile(r"charset=([\w-]+)")
```

`[\w-]+` stops at a quote character. But [RFC 7231 §3.1.1.1](https://www.rfc-editor.org/rfc/rfc7231#section-3.1.1.1) allows the charset value to be a quoted-string, and real servers do send it that way:

```
content-type: text/html; charset="ISO-8859-1"
```

For any quoted charset the regex fails to match, so the method silently falls back to its `utf-8` default. A page actually served as quoted `ISO-8859-1` / `windows-1252` / `Shift_JIS` then gets decoded as UTF-8 → **mojibake**.

```python
>>> ResponseFactory._ResponseFactory__extract_browser_encoding('text/html; charset="ISO-8859-1"')
'utf-8'   # ← wrong, should be 'ISO-8859-1'
```

## Fix

Allow an optional surrounding quote so the value is captured without it:

```python
__CHARSET_RE__ = re_compile(r"""charset=["']?([\w-]+)""")
```

Unquoted headers are completely unaffected; quoted single- or double-quote values are now unwrapped correctly.

## Tests

The `content_type_map` fixture in `tests/fetchers/test_utils.py` (which already lists ~40 quoted-charset expectations) had no test consuming it. Added focused tests for unquoted, quoted, and missing-charset cases:

- The quoted-charset test **fails on `main`** (`utf-8` instead of `ISO-8859-1`) and passes with this fix, confirming it catches the bug.
- Full `tests/fetchers/test_utils.py` passes (the one unrelated failure, `test_generate_headers_browser_mode`, fails identically on unmodified `main` — a `browserforge` env issue, not touched here).
- `ruff check` and `ruff format --check` clean.

Note for maintainers: the `content_type_map` fixture also expects `text/html` / `text/plain` *without* a charset to default to `ISO-8859-1` (the historical default for `text/*`), but the function returns a flat `utf-8` for all charset-less inputs. I kept this PR scoped to the quoting bug; happy to follow up on the `text/*` default separately if you'd like.


D4Vinci:
LGTM, thanks for catching this!

Issue #322, opened 2026-06-02
Bug: Wrong denominator in __are_alike similarity calculation (checks uses candidate_attributes length instead of original_attributes)


kuishou68:
## Bug Description

In `scrapling/parser.py`, the `__are_alike` method has a bug in the similarity score denominator.

### Location
`scrapling/parser.py`, line 994, inside the `__are_alike` method.

### Problem
The similarity score is computed by iterating over `original_attributes.items()` (line 992), which produces at most `len(original_attributes)` ratio values. However, the denominator `checks` is incremented by `len(candidate_attributes)` instead:

```python
if original_attributes:
    score += sum(
        SequenceMatcher(None, v, candidate_attributes.get(k, "")).ratio()
        for k, v in original_attributes.items()
    )
    checks += len(candidate_attributes)  # BUG: should be len(original_attributes)
```

### Impact
When the original element and candidate element have different numbers of attributes, the similarity score is miscalculated:
- **Candidate has more attributes**: score is artificially deflated (dividing by too large a number), causing similar elements to be missed
- **Candidate has fewer attributes**: score is artificially inflated (dividing by too small a number), potentially matching dissimilar elements

### Fix
Change line 994 from:
```python
checks += len(candidate_attributes)
```
to:
```python
checks += len(original_attributes)
```

This ensures the denominator correctly reflects the number of attribute comparisons performed.

D4Vinci:
Hey, thanks for the detailed report, but this is a duplicate of the second bug in #205 , and as I explained there, using `len(candidate_attributes)` is intentional. It penalizes candidates who have extra attributes not present in the original element, making them less "alike".

That said, you are right about one direction here: when the candidate has fewer attributes than the original, the score gets inflated instead of penalized, and that's not the intended behavior. The correct fix for that isn't `len(original_attributes)` though, as that would remove the extra-attributes penalty entirely. It's:

```python
checks += max(len(original_attributes), len(candidate_attributes))
```

This keeps the penalty working in both directions, so I'll fix it that way in the next release. Thanks again!

PR #325, opened 2026-06-02
fix: count original attrs in find_similar threshold

Bortlesboat:
## Problem

`Selector.find_similar()` compares only the original element's attributes, but `__are_alike()` divides that score by `len(candidate_attributes)`. When a candidate has extra attributes, the denominator gets inflated and an otherwise exact attribute match can miss the threshold.

Issue: #322

## Fix

- count `len(original_attributes)` in the similarity denominator
- add a regression test covering a matching sibling that only differs by one extra attribute

## Verification

- `python -m pytest tests/parser/test_find_similar_advanced.py -q`
- `python -m pytest tests/parser/test_general.py -k similar -q`
- `python -m ruff check scrapling/parser.py tests/parser/test_find_similar_advanced.py`
- `python -m ruff format --check scrapling/parser.py tests/parser/test_find_similar_advanced.py`

### DIFF
--- a/scrapling/parser.py
+++ b/scrapling/parser.py
@@ -991,7 +991,7 @@ class Selector(SelectorsGeneration):
                 SequenceMatcher(None, v, candidate_attributes.get(k, "")).ratio()
                 for k, v in original_attributes.items()
             )
-            checks += len(candidate_attributes)
+            checks += len(original_attributes)
         else:
             if not candidate_attributes:
                 # Both don't have attributes, this must mean something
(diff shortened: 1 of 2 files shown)


D4Vinci:
Hey, thanks for the PR, but I can't accept it as it is. As I explained in #205, using `len(candidate_attributes)` here is intentional. It penalizes candidates with extra attributes that the original element doesn't have, and this change removes that penalty entirely. The regression test added here actually asserts the exact behavior I rejected back then (a candidate with an extra attribute scoring a perfect match).

The valid part of #322 is the opposite direction: when the candidate has fewer attributes than the original, the score gets inflated. The correct fix is:
```python
checks += max(len(original_attributes), len(candidate_attributes))
```

which fixes the inflation while keeping the extra-attributes penalty. I will add it right now.

Issue #336, opened 2026-06-03
[BUG] Default follow_redirects=safe breaks all Fetcher.get() calls in 0.4.8


qingquan963:
### Summary
Default `follow_redirects="safe"` (introduced in 0.4.8 as a SSRF-protection feature) breaks **all** `Fetcher.get()` / `Fetcher.post()` / `Fetcher.put()` etc. calls with `ValueError: invalid literal for int() with base 10: 'safe'`.

### Environment
- **Scrapling version:** 0.4.8 (upgraded from 0.4.2)
- **Python:** 3.10
- **OS:** Windows 10
- **curl_cffi:** bundled (latest with Scrapling 0.4.8)

### Reproduction

Minimal example (1 line of code):
```python
from scrapling import Fetcher
f = Fetcher()
r = f.get('https://quotes.toscrape.com/')
```

### Error

```
ValueError: invalid literal for int() with base 10: 'safe'
```

### Full Traceback

```
Traceback (most recent call last):
  File ".../scrapling/fetchers/requests.py", line 33, in get
    return __FetcherClientInstance__.get(url, **_merge_selector_config(cls, kwargs))
  File ".../scrapling/engines/static.py", line 306, in get
    return self._make_request("GET", stealth=stealthy_headers, url=url, **kwargs)
  File ".../scrapling/engines/static.py", line 254, in _make_request
    response = session.request(method, **request_args)
  File ".../curl_cffi/requests/session.py", line 543, in request
    req, buffer, header_buffer, q, header_recved, quit_now = set_curl_options(...)
  File ".../curl_cffi/requests/utils.py", line 525, in set_curl_options
    c.setopt(CurlOpt.FOLLOWLOCATION, int(allow_redirects))
ValueError: invalid literal for int() with base 10: 'safe'
```

### Root Cause

The "safe" mode is meant to be a SSRF-protection value (rejecting redirects to internal/private IPs), but it was introduced as a **string** while `curl_cffi` (the underlying HTTP engine) still expects an **integer**.

Source: `scrapling/engines/static.py:81`
```python
self._default_follow_redirects = kwargs.get("follow_redirects", "safe")
```

This string is later passed all the way down to `curl_cffi/requests/utils.py:525`:
```python
c.setopt(CurlOpt.FOLLOWLOCATION, int(allow_redirects))  # int("safe") raises
```

### Workaround (confirmed working)

Pass `follow_redirects` explicitly as a boolean to skip the "safe" default:
```python
[...]

PR #338, opened 2026-06-03
fix: resolve follow_redirects string modes to curl_cffi-accepted values (#336)

Brentbin:
## What

`Fetcher.get()` (and the other HTTP verbs) raise `ValueError: invalid literal for int()` whenever `follow_redirects` is one of Scrapling's documented string modes. Since the default is `"safe"`, this can break basic fetches depending on the installed curl_cffi version.

## Why

`_ConfigurationLogic._merge_request_args` (scrapling/engines/static.py) forwarded the raw `follow_redirects` value straight to curl_cffi as `allow_redirects`. curl_cffi turns `allow_redirects` into a curl option via `int(allow_redirects)` and only special-cases the literal `"safe"` internally (curl_cffi/requests/utils.py). So `"all"`, `"obeycode"`, and `"firstonly"` hit `int("all")` and raise, and `"safe"` only survives by relying on that internal special case.

## Fix

- Add a module-level `_FOLLOW_REDIRECTS_MAP` + `_resolve_follow_redirects()` in `static.py` that map the four `FollowRedirects` string modes to the matching `CurlFollow` enum member (`from curl_cffi import CurlFollow`).
- Booleans and already-resolved `CurlFollow` values pass through unchanged.
- Unknown strings fall back to `CurlFollow.SAFE` (the SSRF-safe default), so a typo degrades safely rather than crashing — note this replaces the previous loud `ValueError` with a silent fallback.
- Resolve `"safe"` -> `CurlFollow.SAFE` ourselves so behavior no longer depends on curl_cffi's internal special-casing and is robust across versions.

## Tests

Added `TestMergeRequestArgsResolvesFollowRedirects` to `tests/fetchers/test_merge_request_args.py` (default + 4 literals + 2 bool cases). Confirmed RED before the fix (the raw strings leaked through) and GREEN after. Also fed each resolved value through curl_cffi's exact `FOLLOWLOCATION`-setting path; all set without `ValueError`.

## Behavior note
Unknown/typo `follow_redirects` strings now fall back to `CurlFollow.SAFE` instead of raising the previous (incidental) `ValueError: invalid literal for int()`. Since `FollowRedirects` is a `Literal` type, type-checked callers can't hit this; the fallback just makes the runtime robust across curl_cffi versions. Happy to switch to an explicit raise on unknown strings if you prefer fail-loud.

Targeted at `dev` per CONTRIBUTING. `ruff check` + `ruff format --check` clean; `pytest tests/fetchers/test_merge_request_args.py` 12/12 pass on `dev`.

Submitted by @Brentbin; AI-assisted with Claude Code and reviewed before submission.

### DIFF
--- a/scrapling/engines/static.py
+++ b/scrapling/engines/static.py
@@ -4,7 +4,7 @@ from time import sleep as time_sleep
 from asyncio import sleep as asyncio_sleep
 
 from curl_cffi.curl import CurlError
-from curl_cffi import CurlHttpVersion
+from curl_cffi import CurlFollow, CurlHttpVersion
 from curl_cffi.requests import (
     BrowserTypeLiteral,
     Session as CurlSession,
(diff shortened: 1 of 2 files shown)


D4Vinci:
Closed as per #336

PR #340, opened 2026-06-04
fix: prevent IndexError in adaptive relocation when auto_save is enabled

Mubashirrrr:
## Bug

Calling `css()` / `xpath()` with both `adaptive=True` and `auto_save=True` raises `IndexError: list index out of range` whenever the saved element cannot be relocated above the `percentage` threshold (e.g. the page structure changed too much, or a high threshold was set).

### Reproduce

```python
from scrapling import Selector

old = Selector('<div><article id="target"><h3>Widget</h3></article></div>',
               url="example.com", adaptive=True)
new = Selector('<html><body><span>totally unrelated</span></body></html>',
               url="example.com", adaptive=True)

old.css("#target", identifier="target", auto_save=True)

# Nothing on the new page clears the 95% threshold -> relocate() returns []
new.css("#target", identifier="target", adaptive=True, auto_save=True, percentage=95)
# IndexError: list index out of range  (scrapling/parser.py)
```

This is the worst possible moment to crash: it happens precisely when adaptive resilience is supposed to kick in (the structure changed enough that nothing matches above the threshold), and `auto_save` is the documented way to persist the relocated element for faster future lookups.

## Root cause

In `Selector.xpath()`, the relocation branch guards the re-save with:

```python
elements = self.relocate(element_data, percentage)
if elements is not None and auto_save:
    self.save(elements[0], identifier or selector)
```

`relocate()` returns an **empty list** (never `None`) when no candidate clears the threshold, so `elements is not None` is always `True` in this branch and `elements[0]` is evaluated on an empty list.

## Fix

Use a truthiness check so the re-save is skipped when relocation yields nothing:

```python
if elements and auto_save:
    self.save(elements[0], identifier or selector)
```

The successful-relocation path (which re-saves the relocated element) is unchanged — verified that a relocation clearing the threshold still re-saves correctly.

## Test

Added `test_relocation_auto_save_no_match_above_threshold` to `tests/parser/test_adaptive.py`. It is a fail-before / pass-after regression test:

- **Before the fix:** fails with `IndexError: list index out of range`.
- **After the fix:** passes (the call returns an empty `Selectors`).

The full `tests/parser/` suite (93 tests) and `tests/spiders/` suite (347 tests) pass, and `ruff` is clean on the changed files.


D4Vinci:
Wow! Thanks for bringing this up!

PR #342, opened 2026-06-04
docs(proposal): optional firefox stealth engine for StealthyFetcher

feder-cr:
draft / interest-check only, not a feature implementation yet.

discussion: #341

this is the firefox-side version of issue #291 (CloakBrowser stealth backend). three things make it a small lift: executable_path already exists in the fetcher options; camoufox was the StealthyFetcher engine until 0.3.13 (dropped for patchright over memory/stability), so a firefox stealth engine was the original one; and #291/#191/#146 are the recurring pain an engine swap targets.

patchright patches chromium at the js layer (still readable via native .toString() enumeration, the CDP attach signature, chromium-shape risk weighting). a firefox build patched at the c++ source level has no js shim and no CDP signature, and sidesteps the camoufox memory/stability issues since patches live in the compiled binary and it's fully pref-driven.

proposed shape: a firefox branch in the StealthyFetcher engine launching playwright.firefox.launch(executable_path=<patched binary>, firefox_user_prefs=<prefs>) reusing the existing executable_path plumbing. binary is feder-cr/invisible_firefox (MPL-2), auto-downloads on first run. this PR is just the proposal doc.

honest caveats: helps the fingerprint/engine layer only, not IP reputation or a Press & Hold; firefox via playwright has no CDP so the cdp_url path stays chromium-only. if in scope i'll wire the firefox engine branch + docs example. if not, happy to close without noise.

### DIFF
--- a/docs/proposals/invisible-firefox-backend.md
+++ b/docs/proposals/invisible-firefox-backend.md
@@ -1,0 +1,68 @@
+# Stealth patched-Firefox engine option (proposal)
+
+> Status: Draft proposal / interest check
+> Tracking discussion: TBD
+
+## Goal
+
+Offer an optional Firefox engine for `StealthyFetcher`, backed by a Firefox 150
+binary with fingerprint patches applied at the C++ source level, for targets
+where the current Chromium + patchright path still trips Cloudflare, Datadome,
+PerimeterX, or Akamai.
+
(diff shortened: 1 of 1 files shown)


D4Vinci:
There's no changes in this PR

PR #344, opened 2026-06-07
fix(spiders): use os.replace for atomic checkpoint/cache writes on Windows

MrStarkEG:
## Summary

`CheckpointManager.save()` and `ResponseCacheManager.put()` write to a temporary file and then move it into place with `Path.rename()`. On Windows, `os.rename` cannot overwrite an existing destination and raises `FileExistsError` (`WinError 183`), so **every write after the first one fails**:

- **Checkpoint** — the exception propagates, so checkpoint saving (and therefore crawl resume) breaks once a checkpoint already exists.
- **Dev response cache** — the exception is swallowed, the temp file is removed, and `get()` keeps returning the **stale** cached response.

```text
WARNING: Failed to cache response ...: [WinError 183] Cannot create a file
when that file already exists: '...\<fp>.tmp' -> '...\<fp>.json'
```

## Fix

Use `Path.replace()` (`os.replace`) instead of `Path.rename()`. `os.replace` overwrites the destination atomically on every platform. On POSIX, `os.rename` already overwrites, so this is a **no-op on Linux and macOS** — it only changes the broken Windows overwrite case. Verified by running the same temp→final move on both Linux and Windows: identical behaviour on POSIX, fixed on Windows.

## Tests

- Added `test_put_overwrites_existing_entry` covering the cache overwrite path — it fails on Windows before this change (returns the stale body) and passes after. The checkpoint overwrite is already covered by `test_multiple_saves_overwrite`.
- No regressions in the offline test suite.

> Note: the test workflow runs on macOS only, where these tests pass with or without the fix (POSIX `rename` already overwrites); the failure is specific to Windows.


D4Vinci:
Wow, thanks a lot for bringing this up!

PR #351, opened 2026-06-12
fix: handle compound extensions in LinkExtractor

### LINKED ISSUE #349: [BUG] [Good First Issue] LinkExtractor never filters .tar.gz links
## Summary

## Reproduction

```python

```

## Expected

## Suggested fix

axelray-dev:
## Proposed change

`_url_extension()` returns only the last suffix of a URL path segment (e.g. `"gz"` for `dataset.tar.gz`). This means compound extensions like `"tar.gz"` that are already listed in `IGNORED_EXTENSIONS` never match during `deny_extensions` filtering, so archive files such as `.tar.gz` and `.tar.bz2` pass through the extractor undetected.

This PR adds a `_url_extensions()` helper that returns all dot-suffixes for the last path segment (e.g. `{"gz", "tar.gz"}` for `dataset.tar.gz`) and uses it in `_url_passes()` so compound extensions are correctly rejected. The existing `_url_extension()` function is preserved unchanged for backward compatibility.

### Type of change:

- [ ] Dependency upgrade
- [ ] Add or change doctests?
- [ ] Documentation change?

### Additional information

- This PR fixes or closes an issue: fixes #349
- Tests run: `pytest tests/spiders/test_links.py -v` -- 34/34 passed
- Lint: `ruff check` and `ruff format --check` both clean
- Whitespace: `git diff --check` clean

### Checklist:
* [x] I have read CONTRIBUTING.md.
* [x] All function parameters and return values are annotated with Python type hints.

### DIFF
--- a/scrapling/spiders/links.py
+++ b/scrapling/spiders/links.py
@@ -157,6 +157,15 @@ def _url_extension(url: str) -> str:
     return last.rsplit(".", 1)[1].lower()
 
 
+def _url_extensions(url: str) -> set[str]:
+    path = urlsplit(url).path
+    _, _, last = path.rpartition("/")
+    parts = last.lower().split(".")
+    if len(parts) < 2:
+        return set()
+    return {".".join(parts[i:]) for i in range(1, len(parts))}
+
+
 def _filler(x):
     return x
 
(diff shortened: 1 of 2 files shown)


yetval:
Confirmed this fixes it. `.tar.gz` gets dropped by default now, and explicit `deny_extensions={"tar.gz"}` works too while plain `.gz` is still kept. Tests cover the cases well.

Only nit: `_url_extension` (singular) has no callers after this, so it could be dropped, but no big deal. Thanks for the fix!

D4Vinci:
Thanks for tackling this! Closing as a duplicate of #349's fix. We're going with #359, which keeps `_url_extension()` intact for backward compatibility and handles the trailing-dot and double-dot edge cases a bit more cleanly.

Really appreciate the contribution and hope to see more from you. 🙏

PR #353, opened 2026-06-14
docs: clarify stealth init script visibility

### LINKED ISSUE #350: No context persistence for code loaded using init_script

3.14.4

0.4.9

scrapling==0.4.9
playwright==1.60.0
patchright==1.60.1
browserforge==1.2.4
curl_cffi==0.15.0
anyio==4.13.0
Protego==0.6.0
lxml==6.1.1
tld==0.13.2
cssselect==1.4.0
orjson==3.11.9
w3lib==2.4.1
msgspec==0.21.1
typing_extensions==4.15.0

macOS Tahoe 26.5.1

Yes

### Expected behavior

When an absolute file path is provided to the `init_script` parameter, the specified JavaScript code should be registered globally in the browser context, in a similar way to Playwright's `browser_context.add_init_script()`. This script must execute on the creation of every new document / page and survive across page navigations. Any global variables, objects or functions initialised by this script must remain fully accessible within the `page_action` automation hook after the completion of `session.fetch()`.

### Actual behavior

If a JavaScript file path is supplied to the `init_script` parameter, the initialisation state is lost when `session.fetch()` is executed. Consequently, by the time the `page_setup` / `page_action` callback executes, any global variables, functions or objects initialised by the script at the start of the document are missing entirely (evaluating to None or undefined). It appears that the script execution context is dropped or wiped entirely during internal page navigation or context switching. Likewise, using `page.add_init_script()` to load js code within `page_setup` callback also results in the same loss of context.

### Steps To Reproduce

To reproduce this behavior we can simply check if a initialized function is accessible on page_action callback. The same can be done using page_setup which results in the same behavior. Here is a mininimal reproducable example:

```python
import asyncio
import tempfile

MINIMAL_JS_CODE = """
window.calculateSum = function(a, b) {
    return a + b;
};
console.log("[Browser Context] Registered window.calculateSum successfully!");
"""


async def check_execution_context(page):
    print("\n--- Checking Execution Context ---")

    # Attempt to execute the function initialized by the init_script
    try:
        result = await page.evaluate("() => window.calculateSum(1, 1)")
        print(f"Result of window.calculateSum(1, 1): {result}")
    except Exception as e:
        print(f"Execution Error: {e}")
        result = None

    print("\n--- Assertion Results ---")
    if result == 2:
        print("✅ SUCCESS: The initialization context survived page navigation!")
    else:
        print("❌ FAILED: The loaded JS context was wiped out during navigation.")
    print("-------------------------------------------------------\n")


async def process():

    with tempfile.NamedTemporaryFile(mode="w", suffix=".js", delete=False) as temp_js:
        temp_js.write(MINIMAL_JS_CODE)
        temp_js_path = temp_js.name

    async with AsyncStealthySession(
        init_script=temp_js_path,
        page_action=check_execution_context,
    ) as session:
        await session.fetch("https://example.com")


if __name__ == "__main__":
    asyncio.run(process())
```

Executing this results in the following logs:

```
--- Checking Execution Context ---
Execution Error: Page.evaluate: TypeError: window.calculateSum is not a function
    at eval (eval at evaluate (:302:30), <anonymous>:1:14)
    at UtilityScript.evaluate (<anonymous>:309:18)
    at UtilityScript.<anonymous> (<anonymous>:1:44)

--- Assertion Results ---
❌ FAILED: The loaded JS context was wiped out during navigation.
-------------------------------------------------------

[2026-06-12 00:36:55] INFO: Fetched (200) <GET https://example.com/> (referer: https://www.google.com/)
```

mturac:
## Summary
- clarify how `init_script` interacts with Patchright's isolated execution context in Stealthy mode
- point `page_action` users to `page.evaluate(..., isolated_context=False)` when they need to read globals placed on `window`
- keep the existing runtime behavior unchanged

Refs #350

## Tests
- `git diff --check origin/main...HEAD`


yetval:
Thanks for this. I think this is patchright working as intended rather than a bug, since the init script does run, it just sits in the isolated world so your evaluate calls can't see it. Anyone who needs it in the main world can already pass isolated_context=False. My honest take is this doesn't really need a code change and a short docs note would cover it, but happy to be wrong.

mturac:
Good point, thanks for the context. I replaced the runtime change with a docs-only note that keeps the existing behavior intact and points `page_action` users to `page.evaluate(..., isolated_context=False)` when they need to read globals from the page main world.

D4Vinci:
Thanks @yetval @mturac

PR #356, opened 2026-06-17
feat: allow configuring browser navigation wait state

Disaster-Terminator:
## Summary
- Add `wait_until` for browser navigation and pass it to Playwright `page.goto(...)`.
- Support DynamicFetcher and StealthyFetcher, including sync/async, one-shot, session-backed, CLI, and MCP paths.
- Document the interaction with Scrapling's separate `network_idle` wait.

## Compatibility
Default behavior is unchanged: `wait_until` defaults to `"load"`, and CLI calls omit it from fetcher kwargs unless the user passes a non-default value.

## Tests
- Validator coverage for default, valid, and invalid `wait_until` values.
- CLI coverage for explicit forwarding and omitted-default behavior.
- MCP coverage for tool exposure and forwarding through one-shot, existing-session, open-session, and screenshot paths.


D4Vinci:
We are already doing that through specific parameters for them

PR #357, opened 2026-06-19
fix: LinkExtractor never filters compound extensions like .tar.gz

### LINKED ISSUE #349: [BUG] [Good First Issue] LinkExtractor never filters .tar.gz links

MrStarkEG:
Fixes #349.

`_url_extension()` returned only the suffix after the last dot, so `dataset.tar.gz` yielded `"gz"` — which is not in `IGNORED_EXTENSIONS` (only `"tar.gz"` is). The deny check at `links.py` never matched, so `.tar.gz` links were kept and passing `deny_extensions=["tar.gz"]` explicitly was silently ignored too.

**Fix:** `_url_extension()` now returns the set of every dot-suffix of the last path segment (`{"gz", "tar.gz"}`), and the filter rejects the URL when any suffix intersects `deny_extensions`. Single-part extensions are unchanged.

Added 4 regression tests: default compound drops, custom compound deny (plain `.gz` still passes), case-insensitivity, and trailing/double-dot edge cases (`file.`, `a..zip`).

**Note — unrelated CI fix (separate commit `afbb0a7`):** `tests/spiders/test_robotstxt.py` expected `Request-rate: 1/10` → `(1, 1)`. Protego 0.6.1 corrected that parse to `(1, 10)`, which broke the test on CI independently of this PR. Updated the expectation to `(1, 10)` (also latently broken on `main`).


D4Vinci:
Hi, thanks for your contribution, but this PR was made against `main` instead of `dev` and includes extra unrelated changes, while there are PRs with much shorter, cleaner solutions.

PR #358, opened 2026-06-22
fix: keep in-flight requests in checkpoints so resume does not lose them

yetval:
## Problem

Periodic checkpoints silently dropped any request that was in-flight when the checkpoint fired. On resume those requests were filtered as duplicates and never fetched, losing them and everything they would have discovered. This hits the default resumable path (crawldir set, interval 300s, concurrency > 1).

## Cause

`dequeue()` removed the request from `_pending` but left it in `_seen`, and `snapshot()` saves only `_pending`. The periodic checkpoint fires with no quiesce, so an in-flight request is excluded from the save yet stays seen, and resume skips `start_requests()`, so it is never re-enqueued.

## Fix

`_pending` now holds every request until it completes (keyed by `id`): `dequeue()` no longer pops, the engine calls the new `Scheduler.complete()` from `_task_wrapper`, and `snapshot()` (unchanged) now captures in-flight work. Resume re-fetches it (at-least-once).

Tests in `tests/spiders/test_scheduler.py` cover tracking, `complete()`, and the snapshot to restore regression; two existing tests were updated to call `complete()`.


yetval:
Heads up: the failing CI check is unrelated to this PR. All four jobs fail only on `spiders/test_robotstxt.py::TestGetDelayDirectives::test_returns_request_rate_when_set` (201 passed, including the scheduler/checkpoint tests).

The fixture declares `Request-rate: 1/10` but the test asserts the parsed result equals `(1, 1)`. A newer `urllib.robotparser` in the CI runners now correctly parses it as `(1, 10)`, so the stale assertion fails. This PR only touches `scheduler.py`, `engine.py`, and `test_scheduler.py`.

Happy to fix that assertion here or in a separate PR if you prefer.

samrusani:
I ran the focused checks locally on Python 3.13.12:

```text
./.venv313/bin/python -m pytest tests/spiders/test_scheduler.py -q
25 passed

./.venv313/bin/python -m pytest tests/spiders/test_engine.py -q
67 passed

./.venv313/bin/python -m pytest tests/spiders/test_checkpoint.py tests/spiders/test_force_stop_checkpoint.py -q
30 passed
```

The main fix looks coherent, but I found one narrow edge case from changing `_pending` to be keyed by `id(request)`: if the exact same `Request` instance is enqueued more than once with `dont_filter=True`, the queue has multiple entries but the checkpoint mirror collapses them to one.

Repro:

```py
scheduler = Scheduler()
request = Request("https://example.com/retry", dont_filter=True)
await scheduler.enqueue(request)
await scheduler.enqueue(request)

assert len(scheduler) == 2
requests, _ = scheduler.snapshot()
assert len(requests) == 2  # currently 1

first = await scheduler.dequeue()
scheduler.complete(first)
requests, _ = scheduler.snapshot()
assert len(requests) == 1  # currently 0, so the second queued copy is lost from checkpoints
```

This seems contrary to the new invariant in the comment (`every enqueued, not-yet-completed request`). One possible fix is to keep `_pending` keyed by the queue counter, and have `complete(request)` remove one pending entry whose request object is the completed request, rather than keying the mirror by object id.

yetval:
Thank you so much @samrusani, this is a really sharp catch and I appreciate you taking the time to write the repro. You're right that the same request enqueued twice was getting collapsed and lost from the checkpoint. I've reworked the tracking so each copy is kept and removed on its own, and added a test for exactly this case. Focused suites all pass. Thanks again for the careful review.

D4Vinci:
Thanks a lot, @yetval, for noticing and fixing this one!

PR #359, opened 2026-06-26
fix: honor compound link deny extensions

### LINKED ISSUE #349: [BUG] [Good First Issue] LinkExtractor never filters .tar.gz links

renbkna:
Fixes #349.

## What changed
Link filtering now checks all suffixes from the final URL path segment when applying `deny_extensions`, so compound extensions like `.tar.gz` can be denied while preserving exact `.gz` behavior.

## Testing
- Reproduced the original failure before the fix: `.tar.gz` was in `IGNORED_EXTENSIONS`, but `LinkExtractor().matches(...)` returned `True`.
- Verified after the fix: default and explicit `deny_extensions={"tar.gz"}` both reject `.tar.gz`, while plain `.gz` is still allowed when only `tar.gz` is denied.
- `.venv/bin/python -m pytest tests/spiders/test_links.py -q`
- `.venv/bin/python -m ruff check scrapling/spiders/links.py tests/spiders/test_links.py`
- `.venv/bin/python -m ruff format --check scrapling/spiders/links.py tests/spiders/test_links.py --diff`
- `git diff --check`

### DIFF
--- a/scrapling/spiders/links.py
+++ b/scrapling/spiders/links.py
@@ -150,11 +150,17 @@ def _compile_patterns(patterns: Union[str, Pattern[str], PatternInput, None]) ->
 
 
 def _url_extension(url: str) -> str:
+    extensions = _url_extensions(url)
+    return extensions[0] if extensions else ""
+
+
+def _url_extensions(url: str) -> Tuple[str, ...]:
     path = urlsplit(url).path
     _, _, last = path.rpartition("/")
     if "." not in last:
-        return ""
-    return last.rsplit(".", 1)[1].lower()
+        return ()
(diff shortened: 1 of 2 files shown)


renbkna:
I noticed #357 after opening this. This PR targets `dev` per CONTRIBUTING.md and keeps the fix focused with regression tests. Happy to close if maintainers prefer to continue with #357.

D4Vinci:
Thanks for solving this! We will go with your solution instead of the other two PRs, since it's the cleanest and covers all edge cases.

PR #361, opened 2026-06-28
fix: keep init scripts available to page actions

### LINKED ISSUE #350: No context persistence for code loaded using init_script

malsabbagh05:
## Summary
- preserve normal context-level `init_script` registration
- add a marker to detect when the browser backend did not run the init script
- apply the configured init script before `page_action` only when the marker is missing
- add an async stealth regression test for issue #350

## Tests
```bash
python -m compileall -q scrapling/engines/_browsers tests/fetchers/async/test_stealth_session.py
pytest tests/fetchers/async/test_stealth_session.py tests/fetchers/sync/test_stealth_session.py tests/fetchers/async/test_dynamic_session.py tests/fetchers/sync/test_dynamic.py -q
```

Result:

```text
19 passed
```

Also verified the original issue shape against `https://example.com`, where `page_action` observed `seen [2]`.

Fixes #350

### DIFF
--- a/scrapling/engines/_browsers/_base.py
+++ b/scrapling/engines/_browsers/_base.py
@@ -1,7 +1,9 @@
 from time import time
 from re import search as re_search
+from json import dumps as json_dumps
 from asyncio import sleep as asyncio_sleep, Lock
 from contextlib import contextmanager, asynccontextmanager
+from pathlib import Path
 
 from playwright.sync_api._generated import Page
 from playwright.sync_api import (
@@ -43,6 +45,41 @@ from scrapling.core._types import (
     AsyncGenerator,
 )
 from scrapling.engines.constants import STEALTH_ARGS, HARMFUL_ARGS, DEFAULT_ARGS
+
+
+__INIT_SCRIPT_MARKER__ = "__scrapling_init_scripts__"
+
+
+def _get_init_script_key(init_script: str) -> str:
+    """Return a stable key for an init script path."""
+    return str(Path(init_script).resolve())
+
+
+def _get_init_script_source(init_script: str) -> tuple[str, str]:
+    """Return a stable key and source for an init script path."""
+    path = Path(init_script).resolve()
+    return str(path), path.read_text(encoding="utf-8")
+
+
+def _get_init_script_marker(init_script: str) -> str:
+    """Return a marker script used to detect whether context init scripts ran."""
+    key = _get_init_script_key(init_script)
+    return (
+        f"globalThis.{__INIT_SCRIPT_MARKER__} = globalThis.{__INIT_SCRIPT_MARKER__} || {{}};\n"
+        f"globalThis.{__INIT_SCRIPT_MARKER__}[{json_dumps(key)}] = true;"
+    )
+
+
+__INIT_SCRIPT_FALLBACK__ = f"""
+(payload) => {{
+    const [key, source] = payload;
+    globalThis.{__INIT_SCRIPT_MARKER__} = globalThis.{__INIT_SCRIPT_MARKER__} || {{}};
+    if (!globalThis.{__INIT_SCRIPT_MARKER__}[key]) {{
+        (0, eval)(source);
+        globalThis.{__INIT_SCRIPT_MARKER__}[key] = true;
+    }}
+}}
+"""
 
 
 class SyncSession:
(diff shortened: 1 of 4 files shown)


samrusani:
Ran a local review/validation pass against commit `84996b7`.

Validation:

```bash
python -m compileall -q scrapling/engines/_browsers tests/fetchers/async/test_stealth_session.py
python -m pytest tests/fetchers/async/test_stealth_session.py::TestAsyncStealthySession::test_init_script_is_available_in_page_action -q
python -m pytest tests/fetchers/async/test_stealth_session.py tests/fetchers/sync/test_stealth_session.py tests/fetchers/async/test_dynamic_session.py tests/fetchers/sync/test_dynamic.py -q
python -m ruff check scrapling/engines/_browsers/_base.py scrapling/engines/_browsers/_controllers.py scrapling/engines/_browsers/_stealth.py tests/fetchers/async/test_stealth_session.py
git diff --check origin/dev...HEAD
```

Results: targeted regression passed (`1 passed`), the listed browser-session suite passed (`19 passed`), ruff passed, and diff whitespace checks were clean.

I also checked the control flow around `page_setup`/navigation/`page_action`; the fallback stays scoped to the configured `init_script` path and still preserves the normal context-level registration path. I did not find a blocking issue in this patch.

yetval:
Thanks, but this was already settled in #353: the init script does run, it just sits in a different world than `page.evaluate`, so `isolated_context=False` is the whole fix. This patch only re-runs the script after load, which masks the symptom. I'd close in favor of #353.

samrusani:
Thanks for clarifying. That makes sense: my local pass only validated that this patch made the `page_action` repro observe the global, not that re-running the script after load was the right fix. I agree the `isolated_context=False` / docs path in #353 is the cleaner resolution.

malsabbagh05:
Thanks for the clarification. I agree that the #353 approach is the cleaner fix here, since the issue is the script world/visibility behavior rather than needing to re-run the init script after load. Closing this in favor of #353.

PR #365, opened 2026-07-03
Fix Playwright page content encoding

### LINKED ISSUE #364: Browser fetchers: header charset applied to page.content() bytes that are always UTF-8 → mojibake on non-UTF-8 sites
## Bug

When a site serves a non-UTF-8 charset in its `Content-Type` header (e.g. `text/html; charset=iso-8859-15`), every browser-based fetch returns mojibake in `response.html_content` (`ü` → `ÃŒ`, `ö` → `Ãķ`, ...). Verified on scrapling 0.4.8 with `AsyncStealthySession` against https://www.vlotho.de/ (a real municipality site serving `charset=iso-8859-15`).

## Root cause

In `scrapling/engines/toolbelt/convertor.py`, `ResponseFactory.from_async_playwright_response` (and the sync variant):

1. `encoding` is extracted from the HTTP header → `"iso-8859-15"`.
2. But the content bytes come from the live DOM: `page_content = (await cls._get_async_page_content(page)).encode("utf-8")`. `page.content()` is already a correctly decoded Unicode str (the browser handled the wire charset), so these bytes are **always UTF-8** regardless of the header.
3. The mismatched pair is passed to `Response(content=<utf-8 bytes>, encoding="iso-8859-15")`, so `parser.py` builds `lxml.HTMLParser(encoding="iso-8859-15")` and parses UTF-8 bytes as Latin-9 — every multi-byte UTF-8 sequence shreds into two chars (`ü` = `C3 BC` → `Ã` + `Œ`).
4. `html_content` then re-serializes/decodes with the same wrong codec, baking the mojibake into the returned string.

The header encoding is only correct for the fallback branch that uses `final_response.body()` (raw wire bytes). The `page.content()` branch always produces UTF-8.

## Fix

In both `from_playwright_response` and `from_async_playwright_response`: when content came from `page.content()`, pass `encoding="utf-8"` instead of the header-derived encoding (keep the header encoding only for the `final_response.body()` fallback).

## Repro

```python
import asyncio

async def main():
    async with AsyncStealthySession(max_pages=1, headless=True) as session:
        response = await session.fetch("https://www.vlotho.de/")
        print(response.encoding)  # iso-8859-15
        print("BÃŒrger" in response.html_content)  # True — should be "Bürger"

asyncio.run(main())
```

ymxlx:
## Summary

- Set browser `page.content()` responses to `utf-8` after encoding the browser-decoded Unicode HTML.
- Keep the server-provided `Content-Type` charset for raw `final_response.body()` fallback responses.
- Add sync and async regression coverage for non-UTF-8 headers with accented text.

Fixes #364.

## Tests

- `.venv/bin/python -m pytest tests/fetchers/test_response_handling.py`
- `.venv/bin/python -m pytest tests/parser/test_general.py`

### DIFF
--- a/scrapling/engines/toolbelt/convertor.py
+++ b/scrapling/engines/toolbelt/convertor.py
@@ -122,6 +122,7 @@ class ResponseFactory:
         try:
             if page and "html" in final_response.all_headers().get("content-type", ""):
                 page_content = cls._get_page_content(page).encode("utf-8")
+                encoding = "utf-8"
             else:
                 page_content = final_response.body()
         except Exception as e:  # pragma: no cover
@@ -269,6 +270,7 @@ class ResponseFactory:
         try:
             if page and "html" in (await final_response.all_headers()).get("content-type", ""):
                 page_content = (await cls._get_async_page_content(page)).encode("utf-8")
+                encoding = "utf-8"
             else:
                 page_content = await final_response.body()
         except Exception as e:  # pragma: no cover
--- a/tests/fetchers/test_response_handling.py
+++ b/tests/fetchers/test_response_handling.py
@@ -1,7 +1,55 @@
-from unittest.mock import Mock
+from unittest.mock import AsyncMock, Mock
+
+import pytest
 
 from scrapling.parser import Selector
 from scrapling.engines.toolbelt.convertor import ResponseFactory, Response
+
+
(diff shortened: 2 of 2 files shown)


D4Vinci:
LGTM, thanks!

PR #368, opened 2026-07-06
fix(ai): strip control characters before extraction (#366)

### LINKED ISSUE #366: MCP server 'get' tool fails on pages with control characters (e.g. U+0008)
## Problem

The MCP server `get` (and `fetch`, `bulk-get`, etc.) tool fails on pages whose content contains ASCII control characters (e.g. `U+0008` backspace) with the error:

```
Error executing tool get: All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters
```

## Reproduction

```bash
mcp2cli @scrapling get --url "https://jfinal.com/doc/1-5" --main-content-only
```

Also reproducible with:
- https://jfinal.com/doc/1-6
- https://jfinal.com/doc/2-2
- https://jfinal.com/doc/3-7
- https://jfinal.com/doc/3-12
- https://jfinal.com/doc/10-2
- https://jfinal.com/doc/14-5
- https://jfinal.com/doc/14-6

## Root Cause

The page content contains a literal `U+0008` (backspace) control character. The MCP server's response serialization cannot handle this character.

Confirmed via curl:
```python
curl -s "https://jfinal.com/doc/1-5" | python3 -c "
import sys
html = sys.stdin.buffer.read().decode('utf-8', errors='replace')
bad = [(i, hex(ord(c))) for i, c in enumerate(html) if 0 < ord(c) < 32 and c not in '\n\r\t']
print(f'Control chars: {bad}')
"
# Output: Control chars: [(13930, '0x8')]
```

## Impact

8 out of 99 JFinal documentation pages are inaccessible via the MCP server. Any real-world web page with control characters (backspace, null bytes, etc.) in its source will trigger this error, making the tool unreliable for general crawling.

## Suggested Fix

In `scrapling/core/ai.py`, the `_translate_response()` function should sanitize control characters from page content before building the ResponseModel:

```python
import re

def _sanitize(text: str) -> str:
    # Remove ASCII control characters except tab/newline/carriage return
    return re.sub(r"[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]", "", text)
```

Then apply `_sanitize()` to each content string in `_translate_response()`.

## Environment

- scrapling: 0.4.9 (PyPI)
- MCP server run via `uvx scrapling mcp`
- Tested via `mcp2cli @scrapling`

yetval:
Fixes #366.

Pages with control chars like U+0008 crash the MCP `get`/`fetch` tools. This strips those chars (keeping tabs/newlines) so extraction succeeds.

```
Before: ValueError: All strings must be XML compatible...
After:  HTTP 200, content='HelloWorld'
```

Verified by @chcodex on the live JFinal pages. Includes a regression test.


yetval:
Code Quality failure is pre-existing/environmental, seems like it's upstream in the Scrapy module:
```
scrapling/integrations/scrapy.py:15 - error: Import "scrapy.http" could not be resolved
```

D4Vinci:
Thanks @yetval for fixing this.

PR #377, opened 2026-07-15
fix: remove dead fetchall() after INSERT and correct retrieve() error message

fix2015:
- Remove unnecessary cursor.fetchall() call after INSERT OR REPLACE in storage.py (INSERT returns no rows)
- Fix NotImplementedError message in retrieve() abstract method that incorrectly said 'implement save method'

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,11 +98,11 @@ MySpider().start()
   </tr>
   <tr>
     <td width="200">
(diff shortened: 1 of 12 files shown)


D4Vinci:
I don't know how we left that line there, but thanks!

PR #379, opened 2026-07-21
fix(spiders): preserve cookies for browser-engine responses in cache

### LINKED ISSUE #376: ResponseCacheManager drops all cookies for browser-engine responses (cache replay loses cookies)
## Summary

`ResponseCacheManager` silently discards **all** cookies when caching a response produced by a browser (Playwright) engine. On cache replay the restored `Response` has `cookies={}`, so any downstream logic relying on cookies (sessions, auth, CSRF tokens) breaks silently.

## Root cause

For the static engine, `Response.cookies` is a flat `dict` (`scrapling/engines/toolbelt/convertor.py:317` → `dict(response.cookies)`).

For the browser engines it is a `tuple` of **full** cookie dicts:

- `convertor.py:139` (sync): `"cookies": tuple(dict(cookie) for cookie in page.context.cookies()) if page else {}`
- `convertor.py:287` (async): same pattern with `await page.context.cookies()`

In `scrapling/spiders/cache.py:58` the serialization is:

```python
"cookies": dict(response.cookies) if isinstance(response.cookies, dict) else {},
```

Because a browser-engine response has a `tuple`, the `else {}` branch is taken and every cookie is dropped. The round-trip `get()` then rebuilds the `Response` with `cookies={}` (`cache.py:37`), so the cached replay has no cookies at all.

## Steps to reproduce

1. Use a `Page` / `AsyncPage` engine (`fetch` with one of the Playwright backends) to request a URL that sets cookies.
2. Enable response caching (`ResponseCacheManager` / `cache` dir).
3. Replay the cached response.
4. Inspect `response.cookies`.

## Expected

Cookies are preserved across the cache write/read round-trip for browser-engine responses, the same as for static-engine responses.

## Actual

`response.cookies` is `{}` after replay — all cookies are lost.

## Suggested fix

Normalize cookies to a single serializable shape before writing and restore it on read, e.g. always store a list of cookie dicts (`list(response.cookies)` when it is a tuple, or `[response.cookies]` / flatten when it is a flat dict) and pass that back into `Response(cookies=...)` in `get()`.

amitvijapur:
Fixes #376

## Root cause

`Response.cookies` has two possible shapes (`scrapling/engines/toolbelt/custom.py:48`):

- Static engine: a flat `dict` (`convertor.py:317`)
- Browser engines, sync and async (Playwright): a `tuple` of full cookie dicts (`convertor.py:139`, `convertor.py:287`)

`ResponseCacheManager.put()` only handled the `dict` case:

```python
"cookies": dict(response.cookies) if isinstance(response.cookies, dict) else {},
```

Any browser-engine response has `tuple` cookies, so `isinstance(..., dict)` is `False` and the `else {}` branch discards every cookie before it's ever written to the cache file. `get()` then rebuilds the `Response` with `cookies={}`, so a replayed browser-engine response has no cookies at all — this silently breaks anything downstream relying on cookies (sessions, auth, CSRF tokens).

## Fix

Preserve whichever shape the cookies are in, instead of collapsing non-dict cookies to `{}`:

- `put()`: serialize a `tuple` as a JSON array (`list(response.cookies)`) and a `dict` as a JSON object (`dict(response.cookies)`).
- `get()`: JSON arrays deserialize back as `list`, so restore the `tuple` shape `Response.__init__` expects; JSON objects already come back as `dict`.

This keeps the round-trip shape-preserving rather than normalizing everything to a flat dict, so code that depends on the tuple-of-full-cookie-dicts shape for browser-engine responses (domain, path, expiry, etc.) still gets that shape back after a cache replay.

## Tests

Added two regression tests to `tests/spiders/test_cache.py`:

- `test_put_get_roundtrip_preserves_browser_engine_cookies` — tuple-of-dicts cookies (the browser-engine shape) survive `put()`/`get()` unchanged. This fails on `main`/`dev` before the fix (`assert {} == (...)`) and passes after.
- `test_put_get_roundtrip_preserves_static_engine_cookies` — flat-dict cookies (the static-engine shape) still round-trip correctly, confirming no regression for the existing working case.

```
tests/spiders/test_cache.py::TestResponseCacheManager::test_put_get_roundtrip_preserves_browser_engine_cookies PASSED
tests/spiders/test_cache.py::TestResponseCacheManager::test_put_get_roundtrip_preserves_static_engine_cookies PASSED
```

Full `tests/spiders/test_cache.py` suite (12 tests) passes. `ruff check`, `ruff format --check`, `mypy`, `pyright`, and `bandit` all pass clean on the changed file. Ran the broader non-browser test suite (`pytest tests/ -k "not (DynamicFetcher or StealthyFetcher)"`, 738 passed) to confirm no regressions elsewhere — the only failures were pre-existing, unrelated tests failing due to missing local Playwright browser binaries.

Credit to @truongsontung for the initial diagnosis in the issue thread — I independently confirmed the same root cause and landed on a slightly different fix (shape-preserving round-trip rather than always flattening to a dict on read) to avoid changing the cookie shape browser-engine consumers see after a cache replay.

### DIFF
--- a/scrapling/spiders/cache.py
+++ b/scrapling/spiders/cache.py
@@ -6,7 +6,7 @@ import anyio
 from anyio import Path as AsyncPath
 
 from scrapling.core.utils import log
-from scrapling.core._types import Dict, Optional, Any
+from scrapling.core._types import Dict, Optional, Any, Tuple
 from scrapling.engines.toolbelt.custom import Response
 
 
@@ -28,13 +28,21 @@ class ResponseCacheManager:
             async with await anyio.open_file(path, "rb") as f:
                 data: Dict[str, Any] = orjson.loads(await f.read())
 
+            # Browser-engine cookies are cached as a JSON array (see `put`) and come back
+            # as a `list`; restore the `tuple` shape `Response` expects. Static-engine
+            # cookies are cached as a JSON object and come back as a `dict` already.
+            cached_cookies = data["cookies"]
+            cookies: Tuple[Dict[str, str], ...] | Dict[str, str] = (
+                tuple(cached_cookies) if isinstance(cached_cookies, list) else cached_cookies
+            )
+
             return Response(
                 url=data["url"],
                 content=b64decode(data["content"]),
                 status=data["status"],
                 reason=data["reason"],
                 encoding=data["encoding"],
-                cookies=data["cookies"],
+                cookies=cookies,
                 headers=data["headers"],
                 request_headers=data["request_headers"],
                 method=data["method"],
@@ -55,7 +63,13 @@ class ResponseCacheManager:
                     "status": response.status,
                     "reason": response.reason,
                     "encoding": response.encoding,
-                    "cookies": dict(response.cookies) if isinstance(response.cookies, dict) else {},
+                    # Browser-engine responses store cookies as a `tuple` of full cookie
+                    # dicts; static-engine responses store a flat `dict`. Preserve whichever
+                    # shape it is instead of collapsing non-dict cookies to `{}`, which was
+                    # silently dropping every cookie from browser-engine responses.
+                    "cookies": list(response.cookies)
+                    if isinstance(response.cookies, tuple)
+                    else dict(response.cookies),
                     "headers": dict(response.headers),
                     "request_headers": dict(response.request_headers),
                     "method": method,
--- a/tests/spiders/test_cache.py
+++ b/tests/spiders/test_cache.py
@@ -14,14 +14,19 @@ from scrapling.engines.toolbelt.custom import Response
 from scrapling.core._types import Any, Dict, Set, AsyncGenerator
 
 
-def _make_response(url: str = "https://example.com", body: bytes = b"<html>hello</html>", status: int = 200) -> Response:
+def _make_response(
+    url: str = "https://example.com",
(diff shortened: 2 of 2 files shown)


yetval:
Reviewed and verified this locally against dev. LGTM.

The root cause is as described, the new browser cookie test fails before the fix and passes after, and the full test suites, lint, and type checks are all clean. Edge cases the tests don't cover (empty tuple cookies, old cache files, float expiries) also behave.

Minor notes, none blocking: passing `response.cookies` straight to orjson would work too, but the explicit conversion makes the two shapes obvious, so keep it. A tagged schema or helper functions would be overengineering for a fix this narrow. The cookie type annotation doesn't match what browser cookies actually contain, but that's inherited from `Response` and belongs in a separate cleanup.

Shape-preserving is the right call over flattening, which would collide keys and lose data.

amitvijapur:
Thanks for the thorough verification — especially exercising the edge cases beyond the tests (empty tuple cookies, old cache files, float expiries). Agreed on all three notes: keeping the explicit conversion for readability, no schema machinery for a fix this narrow, and the `Response.cookies` annotation being a separate cleanup. If it'd be useful, happy to pick that annotation cleanup up in a follow-up PR.

D4Vinci:
LGTM, thanks for solving this!

PR #382, opened 2026-07-24
fix: default browser locale to system setting

### LINKED ISSUE #381: locale doesn't actually default to the system locale; silently defaults to en-US, causing false-positive geo-mismatch blocks (e.g. Google 429)

Python 3.13.5

0.4.11

anyio==4.14.2
browserforge==1.2.4
certifi==2026.7.22
cffi==2.1.0
click==8.4.2
colorama==0.4.6
cssselect==1.4.0
curl_cffi==0.15.0
greenlet==3.5.4
idna==3.18
lxml==6.1.1
markdown-it-py==4.2.0
mdurl==0.1.2
msgspec==0.21.1
orjson==3.11.9
patchright==1.61.2
pip==26.1.2
playwright==1.61.0
Protego==0.6.2
pycparser==3.0
pyee==13.0.1
Pygments==2.20.0
rich==15.0.0
scrapling==0.4.11
tld==0.13.2
typing_extensions==4.16.0
w3lib==2.4.1

Windows 11

Yes

### Expected behavior

StealthyFetcher.fetch("https://www.google.com/search?q=new test", real_chrome=True, headless=False) to be 200

### Actual behavior

429

https://github.com/user-attachments/assets/f4acea02-53f4-4cc9-964b-2309fde81836

### Steps To Reproduce

The docs for StealthyFetcher state (docs/fetching/stealthy.md, locale row):

"Defaults to the system default locale."

But locale defaults to None in _validators.py (locale: str | None = None), and that None is passed straight through to Playwright's new_context(locale=None, ...). Playwright does not inherit the OS locale in this case; it falls back to its own internal default (en-US), regardless of the host machine's actual region.

Impact: Any user running from a non-US region gets requests sent with Accept-Language: en-US / navigator.language: en-US no matter their real locale, which is a geo-mismatch signal some sites (confirmed with Google Search) use to flag/block traffic with a 429 → /sorry challenge — even with real_chrome=True, headless=False.

Fix suggestion: either (a) auto-detect and pass the actual system locale (e.g. via Python's locale.getlocale()) when the user doesn't set one, so the doc claim is true, or (b) fix the docs to say it defaults to en-US and make that limitation explicit, since silent mismatch is the worse of the two options.

For example, on my Brazilian Portuguese Windows 11,
```
>>> import locale
>>> locale.getlocale()
('pt_BR', 'cp1252')
```

Sanjays2402:
## Proposed change

When no locale is provided, resolve the host locale and normalize underscores to Playwright's BCP 47-style format. Explicit locale values remain unchanged.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

- This PR fixes or closes an issue: fixes #381
- This PR is related to an issue: #381
- Link to documentation pull request: N/A

### Checklist:

### DIFF
--- a/scrapling/engines/_browsers/_validators.py
+++ b/scrapling/engines/_browsers/_validators.py
@@ -1,3 +1,4 @@
+import locale
 from pathlib import Path
 from typing import Annotated
 from functools import lru_cache
@@ -95,6 +96,10 @@ class PlaywrightConfig(Struct, kw_only=True, frozen=False, weakref=True):
 
     def __post_init__(self):  # pragma: no cover
         """Custom validation after msgspec validation"""
+        if self.locale is None:
+            system_locale = locale.getlocale()[0]
+            if system_locale:
+                self.locale = system_locale.replace("_", "-")
         if self.page_action and not callable(self.page_action):
             raise TypeError(f"page_action must be callable, got {type(self.page_action).__name__}")
         if self.page_setup and not callable(self.page_setup):
--- a/tests/fetchers/test_validator.py
+++ b/tests/fetchers/test_validator.py
@@ -1,4 +1,7 @@
+from unittest.mock import patch
+
 import pytest
+
 from scrapling.engines._browsers._validators import (
     validate,
     StealthConfig,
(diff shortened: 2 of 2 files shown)


D4Vinci:
Thanks for the PR, but this is not the correct solution. The correct solution is to fix what made the browser not default to the system locale, which I think I found right now. Also, the locale library returns None in many cases, so it's not a reliable solution here as well.

PR #384, opened 2026-07-24
ci(docker): tag images with the release version on merge

JanRK:
## Proposed change

Published Docker images never receive the release version — only `latest` / `main` (from a manual `workflow_dispatch`) and `pr-N`. The `type=semver` rules in `docker-build.yml` never activate, because the workflow runs on `pull_request: closed` / `workflow_dispatch` — never on a tag ref — so `docker/metadata-action`'s semver logic (which requires `refs/tags/*`) is effectively dead, and released versions (`vX.Y.Z`) are never applied to an image.

Triggering the docker build on a `release` / `push tag` event would not work either: the tag/release is created by `github-actions[bot]` using the default `GITHUB_TOKEN`, and GitHub does not start new workflow runs from `GITHUB_TOKEN`-raised events.

Instead, this PR derives the version from the merged PR title — the same source `release-and-publish.yml` already uses — and passes it to `docker/metadata-action` via `value=`, so the semver rules emit the version on the same merge event that creates the release. Purely additive; manual dispatch is unchanged.

**Changes (all in `.github/workflows/docker-build.yml`):**

- **Extract the version from the merged PR title** in a new step, exposed as `steps.version.outputs.version` (set only on a merged `pull_request` whose title starts with `v`). The title is read through an `env:` variable instead of being inlined into the `run:` script, to avoid shell script injection from an attacker-controlled PR title.
- **Rewire the semver tags** to use that value:
  - `type=semver,pattern={{version}},value=…,enable=…` → `X.Y.Z`
  - `type=semver,pattern={{major}}.{{minor}},value=…,enable=…` → `X.Y`
  - `latest` is added automatically by `docker/metadata-action` for non-pre-release semver.
  - The `{{major}}` tag is intentionally dropped: for a `0.y.z` version it would push a bare `0` tag that slides across every `0.x` release.
  - The `enable=` guards make manual dispatch / non-version merges skip the semver lines cleanly (no warnings).
- **Gate the job** with `if: github.event_name == 'workflow_dispatch' || github.event.pull_request.merged == true`, so it no longer builds/pushes on closed-but-unmerged PRs.
- **Add `id: build`** to the build-push step so the existing "Image digest" step prints the real digest instead of an empty string.

**Resulting tags**

| Event | Before | After |
|---|---|---|
| Merge PR `vX.Y.Z` | `pr-N` only | `X.Y.Z`, `X.Y`, `latest` (+ `pr-N`) |
| `workflow_dispatch` on `main` | `main`, `latest` | unchanged |
| PR closed without merge | builds `pr-N` | job skipped |

**Testing**

- `actionlint` passes clean on the workflow.
- Unit-tested the version-extraction logic against real and synthetic PR titles, including a script-injection payload (confirmed the title is handled as literal data, no command execution).
- Verified the `value=` override, `{{major}}.{{minor}}` output, automatic `latest`, and `enable=` gating against `docker/metadata-action` source and README.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

### Checklist:
* [ ] All new Python files are placed inside an existing directory. 
* [ ] All filenames are in all lowercase characters with no spaces or dashes.
* [ ] All functions and variable names follow Python naming conventions. 

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -19,6 +19,7 @@ env:
 jobs:
   build-and-push:
     runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.pull_request.merged == true }}
     permissions:
       contents: read
       packages: write
@@ -26,6 +27,19 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v6
+
+    - name: Extract release version from PR title
+      id: version
+      if: ${{ github.event_name == 'pull_request' && github.event.pull_request.merged == true }}
+      env:
+        PR_TITLE: ${{ github.event.pull_request.title }}
+      run: |
+        if [[ $PR_TITLE =~ ^v ]]; then
+          echo "version=$PR_TITLE" >> "$GITHUB_OUTPUT"
+          echo "Release version detected from PR title: $PR_TITLE"
+        else
+          echo "PR title '$PR_TITLE' is not a version; skipping semver tags."
+        fi
 
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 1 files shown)


D4Vinci:
LGTM, thanks for catching this and solving it!

PR #387, opened 2026-07-26
docs: add an invisible_playwright engine section to StealthyFetcher

feder-cr:
Adds a section to docs/fetching/stealthy.md documenting invisible_playwright as an engine, right after the Camoufox one and following the same subclass-StealthySession pattern. Docs only, one file, no code changes anywhere in the package.

It is a Firefox patched at the C++ source level driven by stock Playwright, and it exposes public config helpers so the launch options can be built without going through its own launcher, which is what makes the snippet fit your start() flow. I kept _initialize_context and _is_alive in the example since the current start() sets both and the Camoufox snippet above predates them.

I dropped a timezone argument I had first written because timezone_id only appears in docstrings and I could not confirm it on _config, so I left it out rather than put a line in your docs I had not checked. Same reason there are no numbers in here.

For context, PR #342 from me was closed as empty, correctly, it shipped no diff at all. This one is the change that should have been in it.

### DIFF
--- a/docs/fetching/stealthy.md
+++ b/docs/fetching/stealthy.md
@@ -338,6 +338,49 @@ async with AsyncStealthySession(solve_cloudflare=True, headless=True) as session
 
 Enjoy! :)
 
+## Using invisible_playwright as an engine
+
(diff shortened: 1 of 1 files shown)


D4Vinci:
Thanks for the PR, but unfortunately, it looks like you misunderstood that section!

It's not planned to add any other library there in the docs. The Camoufox section was added only because the StealthyFetcher was using Camoufox in the past, so it was added for users who want to keep using Camoufox.

In the future, we might study adding your library to Scrapling, but we have a lot of work to do first before looking into that.

PR #393, opened 2026-07-30
fix(ai): size the MCP bulk page pools within the validator bounds

Yigtwxx:
## Proposed change

The two bulk browser tools of the MCP server size their page pool incorrectly, in opposite directions.

**1. `bulk_fetch` cannot handle more than 50 URLs.** On the session-less path it passes `max_pages=len(urls)` to `AsyncDynamicSession`, but `PagesCount` is `Annotated[int, Meta(ge=1, le=50)]` (`scrapling/engines/_browsers/_validators.py`). Any batch over 50 URLs — or an empty list, which trips the `ge=1` bound — fails validation before a single fetch starts:

```python
import asyncio
from scrapling.core.ai import ScraplingMCPServer

asyncio.run(ScraplingMCPServer().bulk_fetch(urls=["https://example.com"] * 60))
# TypeError: Invalid argument type: Expected `int` <= 50 - at `$.max_pages`
```

**2. `bulk_stealthy_fetch` never sets `max_pages` at all**, so the pool falls back to the default of 1. Every URL in the batch queues behind a single tab, and if the pool does not free up within the 60s wait the fetch raises `TimeoutError` (`scrapling/engines/_browsers/_base.py`). The two bulk tools also behaved inconsistently for the same batch size.

Both call sites now go through a small helper that clamps the pool to the validator's range:

```python
_MAX_POOL_PAGES = 50  # Upper bound of `PagesCount` in scrapling/engines/_browsers/_validators.py


def _page_pool_size(urls: Sequence[str]) -> int:
    """Return a page pool size that covers the batch without leaving the validator's bounds."""
    return min(max(len(urls), 1), _MAX_POOL_PAGES)
```

So a batch is fetched with one page per URL up to 50, and larger batches are processed through a pool of 50 concurrent pages instead of raising. The `:param urls:` docstring line of both tools mentions the 50-page cap, since those docstrings are what the model receives as the MCP tool description.

Tests were added in `tests/ai/test_ai_mcp.py` (`TestBulkPagePool`), reusing the existing fake-session monkeypatch pattern so nothing launches a browser: both bulk paths are checked for 3/4, 60 and 0 URLs, and one test feeds the computed size into a real `AsyncDynamicSession` as a regression guard against the validator bounds drifting apart from `_MAX_POOL_PAGES`.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

### Checklist:


D4Vinci:
Nice, thanks for catching this one!

PR #398, opened 2026-08-04
fix: derive chrome fingerprint version from installed datapack at runtime

### LINKED ISSUE #396: Scrapling 0.4.12 fails to import on Linux (ValueError in browserforge at module load)
## Summary

On Linux, `import scrapling.fetchers` (or anything that pulls in the browser engines) raises `ValueError: No headers based on this input can be generated` **at import time**, before any user code runs. Scrapling is therefore unusable out-of-the-box on Linux servers without a source patch. Windows and macOS are unaffected, which is presumably why this slipped through.

## Environment

- Scrapling `0.4.12` (latest at time of writing), installed via `pip install scrapling[fetchers,dynamic]`
- Python `3.12.3`, Ubuntu `24.04.4 LTS` (reproduces on any Linux)
- `apify-fingerprint-datapoints 0.14.0`, `browserforge 1.2.4`

## Reproduction

```bash
python3.12 -m venv /tmp/sv && /tmp/sv/bin/pip install 'scrapling[fetchers,dynamic]'
/tmp/sv/bin/python -c "from scrapling.fetchers import DynamicFetcher"
```

### Traceback (tail)

```
File ".../scrapling/engines/_browsers/_config_tools.py", line 3, in <module>
    __default_useragent__ = generate_headers(browser_mode=True).get("User-Agent")
File ".../scrapling/engines/toolbelt/fingerprints.py", line 56, in generate_headers
    return HeaderGenerator(browser=browsers, os=os_name, device="desktop").generate()
File ".../browserforge/headers/generator.py", line 230, in _get_headers
    raise ValueError("No headers based on this input can be generated. ...")
ValueError: No headers based on this input can be generated. Please relax or change some of the requirements you specified.
```

## Root cause

`scrapling/engines/toolbelt/fingerprints.py` hardcodes the browser version:

```python
chromium_version = 149
chrome_version = 149
```

and `_config_tools.py` runs `generate_headers(browser_mode=True)` (→ Chrome 149 / Linux / desktop) **at module import**. The bundled `apify-fingerprint-datapoints 0.14.0` has **no Chrome-149 Linux-desktop entry**, so `browserforge` raises.

### Evidence (Chrome-on-Linux-desktop across versions)

```
Chrome 149 linux desktop -> FAIL   ← what 0.4.12 requests
Chrome 148 linux desktop -> FAIL
Chrome 145 linux desktop -> FAIL
Chrome 140 linux desktop -> OK
Chrome 135 linux desktop -> OK
Chrome 130 linux desktop -> OK
Chrome 120 linux desktop -> FAIL
Chrome 110 linux desktop -> FAIL
```

The current datapack window for Chrome-on-Linux-desktop is roughly **v127–v140**; `149` is above it.

## Why it matters

Anyone running Scrapling on a Linux server (the most common scraping host) cannot `import` it at all without patching installed source — which is lost on every upgrade. This blocks the primary deployment target for a scraping library.

## Suggested fix

Don't hardcode a single version. Either:

1. **Derive the version from the installed datapack** (query the highest available Chrome-on-target-OS version at runtime), or
2. **Pin to a known-good version inside the datapack window** (e.g. `140`) until the datapack ships `149`, or
3. **Relax the request and fall back** (drop `min_version`/`max_version`, or broaden the OS) when the exact combo has no data, instead of letting `browserforge` raise at import.

### Minimal patch that unblocks immediately

```diff
# scrapling/engines/toolbelt/fingerprints.py
-chromium_version = 149
-chrome_version = 149
+chromium_version = 140
+chrome_version = 140
```

Happy to open a PR for option (1) or (2) if a maintainer confirms the preferred direction. Thanks for the library!

albatrossflyon-coder:
Fixes #396

## Root cause

The module-level `chromium_version = 149` / `chrome_version = 149` constants in `scrapling/engines/toolbelt/fingerprints.py` were hardcoded above the newest Chrome version the installed `apify-fingerprint-datapoints` release actually ships fingerprint data for (max `143` on `apify-fingerprint-datapoints==0.14.0`). Since `__default_useragent__` is computed at module import time, `browserforge`'s `HeaderGenerator` raises `ValueError: No headers based on this input can be generated` as soon as Scrapling is imported.

This is not platform-specific. The issue was originally reported on Linux and confirmed on macOS in the comments, but the same mismatch reproduces identically on **Windows** — I verified this live before writing the fix.

## Fix

Instead of hardcoding a version number that will drift ahead of the installed datapack on every future Scrapling/browserforge release (as [pointed out in the issue thread](https://github.com/D4Vinci/Scrapling/issues/396#issuecomment-5170915056)), derive the newest supported Chrome version directly from the installed datapack's own `browser-helper-file.json` at import time via a small cached helper, `_max_supported_version()`. This is the same data browserforge's own `HeaderGenerator._load_unique_browsers()` already reads, so it's guaranteed to stay in sync with whatever datapack version is actually installed — no more hardcoded floor to bump on every release.

## Testing

- Reproduced the bug live (`ValueError` on `generate_headers(browser_mode="chrome")`) before the fix, confirmed it's resolved after
- Added two regression tests in `tests/fetchers/test_utils.py`: one asserting the version constants match the datapack's actual max, one asserting header generation in chrome mode no longer raises
- Full existing test suite in `tests/fetchers/test_utils.py` passes (5/5, including the 2 new tests)
- `ruff check` / `ruff format --check`: clean
- `bandit`: no issues
- `mypy` / `pyright`: no issues
- `vermin -t=3.10-`: confirms `3.10`, matching `requires-python`
- Static analysis (semgrep, security-audit + secrets rulesets): 0 findings

## Note

Added a defensive check so `_max_supported_version()` raises a clear error instead of an opaque `max() arg is an empty sequence` if it's ever called with a browser name absent from the datapack (not reachable today since it's only called with `"chrome"`, which is always present, but worth guarding since it's a general-purpose helper).

### DIFF
--- a/scrapling/engines/toolbelt/fingerprints.py
+++ b/scrapling/engines/toolbelt/fingerprints.py
@@ -2,9 +2,11 @@
 Functions related to generating headers and fingerprints generally
 """
 
+import json
 from functools import lru_cache
 from platform import system as platform_system
 
+from apify_fingerprint_datapoints import get_browser_helper_file
 from browserforge.headers import Browser, HeaderGenerator
 from browserforge.headers.generator import SUPPORTED_OPERATING_SYSTEMS
 
@@ -12,9 +14,30 @@ from scrapling.core._types import Dict, Literal, Tuple
 
 __OS_NAME__ = platform_system()
 OSName = Literal["linux", "macos", "windows"]
-# Current versions hardcoded for now (Playwright doesn't allow to know the version of a browser without launching it)
-chromium_version = 149
-chrome_version = 149
+
+
+# The installed `apify-fingerprint-datapoints` release only ships fingerprint data up to a
+# certain browser version. Playwright doesn't expose the real installed browser version without
+# launching it, so we ask the datapack itself what the newest version it actually has data for is,
+# instead of hardcoding a version number that can drift ahead of the installed datapack and make
+# browserforge's HeaderGenerator raise `ValueError: No headers based on this input can be
+# generated` at import time (see https://github.com/D4Vinci/Scrapling/issues/396).
+@lru_cache(1, typed=True)
+def _max_supported_version(browser: str) -> int:
+    """Get the newest major version of `browser` that the installed fingerprint datapack has data for.
+
+    :param browser: Browser name as used in the datapack, e.g. ``"chrome"``.
+    :return: Newest supported major version number.
+    """
+    entries = json.loads(get_browser_helper_file().read_bytes())
+    versions = [int(entry.split("/")[1].split(".")[0]) for entry in entries if entry.startswith(f"{browser}/")]
+    if not versions:
+        raise ValueError(f"The installed fingerprint datapack has no entries for browser {browser!r}")
+    return max(versions)
+
+
+chromium_version = _max_supported_version("chrome")
+chrome_version = chromium_version
 
 
 @lru_cache(1, typed=True)
--- a/tests/fetchers/test_utils.py
+++ b/tests/fetchers/test_utils.py
@@ -8,7 +8,13 @@ from scrapling.engines.toolbelt.navigation import (
     create_async_intercept_handler,
     _is_domain_blocked,
 )
-from scrapling.engines.toolbelt.fingerprints import get_os_name, generate_headers
+from scrapling.engines.toolbelt.fingerprints import (
+    get_os_name,
+    generate_headers,
+    chrome_version,
+    chromium_version,
+    _max_supported_version,
+)
 
 
 @pytest.fixture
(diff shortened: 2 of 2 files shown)


D4Vinci:
Nice idea, buddy, but the issue is that it doesn't take the OS into account. Not all operating systems have the same latest version available. I have a fix for that in the works, but don't let this discourage you from sending more PRs in the future!

albatrossflyon-coder:
Good catch on the OS point — makes sense that the max supported Chrome version isn't uniform across platforms, I hadn't accounted for that. If there's an issue or branch tracking your fix I'd like to follow it, and happy to help test across OSes once it's ready. Thanks for scrapling, it's been solid to build on.

PR #395, opened 2026-08-02
fix: defer browser headers until session startup

SulimanAbdulrazzaq:
## Proposed change

Scrapling currently generates browser headers while importing `_config_tools`. If browserforge cannot produce headers for the detected browser version, that exception prevents unrelated entry points such as `scrapling mcp` from starting.

This change defers default user-agent generation until a headless browser session actually needs it and caches one value per browser mode. Explicit user agents and headed sessions keep their existing behavior. Regression tests verify that importing the configuration module does not call browserforge and that on-demand generation preserves the Chromium/Chrome modes and caching behavior.

Validation:
- `pytest -o addopts="" tests/fetchers/test_browser_config.py -q` (2 passed)
- Ruff check and format check for the three changed files
- mypy and pyright for the two changed source files
- Bandit and Vermin repository checks for the changed files
- real `scrapling mcp` stdio initialization returned a valid JSON-RPC result
- `git diff --check`

The exact browserforge exception was reproduced locally with Scrapling 0.4.12, browserforge 1.2.4, and MCP 1.29.0. A macOS runner was not available, so the platform-independent failure path is additionally covered with a deterministic mock.

### Type of change:

- [ ] Dependency upgrade
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New integration (thank you!)
- [ ] New feature (which adds functionality to an existing integration)
- [ ] Deprecation (breaking change to happen in the future)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Add or change doctests? -- Note: Please avoid changing both code and tests in a single pull request.
- [ ] Documentation change?

### Additional information

- This PR fixes or closes an issue: fixes #394
- This PR is related to an issue: #394
- Link to documentation pull request: **Not applicable**

### Checklist:
* [x] I have read [CONTRIBUTING.md](https://github.com/D4Vinci/Scrapling/blob/main/CONTRIBUTING.md).
* [x] This pull request is all my own work -- I have not plagiarized.
* [x] I know that pull requests will not be merged if they fail the automated tests.
* [x] All new Python files are placed inside an existing directory.
* [x] All filenames are in all lowercase characters with no spaces or dashes.
* [x] All functions and variable names follow Python naming conventions.
* [x] All function parameters and return values are annotated with Python [type hints](https://docs.python.org/3/library/typing.html).
* [x] All functions have doc-strings.

Issue #392, opened 2026-07-30
`scrapling mcp` crashes on startup with mcp 2.0.0: No module named 'mcp.server.fastmcp'


KhaiHuynhVN:
### Have you searched if there an existing issue for this?

- [x] I have searched the existing issues

### Python version (python --version)

Python 3.14.6

### Scrapling version (scrapling.__version__)

0.4.12

### Dependencies version (pip3 freeze)

```
annotated_types==0.8.0
anyio==4.14.2
apify_fingerprint_datapoints==0.13.0
attrs==26.1.0
beautifulsoup4==4.15.0
browserforge==1.2.4
certifi==2026.7.22
cffi==2.1.0
click==8.4.2
colorama==0.4.6
cryptography==49.0.0
cssselect==1.5.0
curl_cffi==0.15.0
greenlet==3.5.4
h11==0.16.0
httpcore2==2.9.1
httpx2==2.9.1
idna==3.18
jsonschema==4.26.0
jsonschema_specifications==2025.9.1
lxml==6.1.1
markdown_it_py==4.2.0
markdownify==1.2.3
mcp==2.0.0
mcp_types==2.0.0
mdurl==0.1.2
msgspec==0.21.1
opentelemetry_api==1.44.0
orjson==3.11.9
patchright==1.61.2
playwright==1.61.0
protego==0.6.2
pycparser==3.0
pydantic==2.13.4
pydantic_core==2.46.4
pyee==13.0.1
pygments==2.20.0
pyjwt==2.13.0
python_multipart==0.0.32
pywin32==312
referencing==0.37.0
rich==15.0.0
rpds_py==2026.6.3
scrapling==0.4.12
six==1.17.0
soupsieve==2.9.1
sse_starlette==3.4.6
starlette==1.3.1
tld==0.13.2
truststore==0.10.4
typing_extensions==4.16.0
typing_inspection==0.4.2
uvicorn==0.52.0
w3lib==2.4.1
```

### What's your operating system?

Windows 10 Home 22H2 (10.0.19045)

### Are you using a separate virtual environment?

Yes

### Expected behavior

`scrapling mcp` starts the stdio MCP server and answers a JSON-RPC `initialize` request, so MCP clients (Claude Code, Cursor, etc.) can connect.

### Actual behavior

As of **mcp 2.0.0** (released 2026-07-28), the process exits immediately on startup:

```
Traceback (most recent call last):
  File "<frozen runpy>", line 203, in _run_module_as_main
[...]

PR #399, opened 2026-08-04
fix(fingerprints): fall back to a usable browser version instead of crashing at import

### LINKED ISSUE #394: MCP server crashes on macOS at import: ValueError "No headers based on this input from browserforge

python3.12

0.4.12

scrapling==0.4.12
mcp==1.29.0
click==8.4.2
browserforge==1.2.4

mac OS  26.5.2

No

### Expected behavior

scrapling mcp starts via stdio, accepts JSON-RPC initialize, and returns available tools list. No browser headers needed at startup.

### Actual behavior

Immediate traceback before accepting input:
File "scrapling/engines/_browsers/_config_tools.py", line 3, in ‹module> _default_useragent. = generate_headers (browser_mode=True) •get ("User-Agent") File "scrapling/engines/toolbelt/fingerprints.py", line 56, in generate_headers return HeaderGenerator (browser=browsers, os=o_name, device="desktop") •generate()
File "browserforge/headers/generator.py", line 230, in _get_headers
raise ValueError("No headers based on this input can be generated. Please relax or change some
of the requirements you specified.")
ValueError: No headers based on this input can be generated.

### Steps To Reproduce

echo '{"jsonipc"; "2.0", "method": "initialize", "params": {"protoco IVersion": "2024-11-05", "capabilities":(), "id":1}}' | uvx -from
"scraplinglai］"--with "mcp>=1.29， <2" scrap1ing ncp

### LINKED ISSUE #396: Scrapling 0.4.12 fails to import on Linux (ValueError in browserforge at module load)

AbyAbyss:
Fixes #394, fixes #396.

## The problem

Every browser fetcher fails to import on a clean install:

```
$ pip install "scrapling[fetchers]"
$ python -c "from scrapling.fetchers import StealthyFetcher"
...
  File "scrapling/engines/_browsers/_config_tools.py", line 3, in <module>
    __default_useragent__ = generate_headers(browser_mode=True).get("User-Agent")
  File "scrapling/engines/toolbelt/fingerprints.py", line 56, in generate_headers
    return HeaderGenerator(browser=browsers, os=os_name, device="desktop").generate()
  File "browserforge/headers/generator.py", line 230, in _get_headers
    raise ValueError(
ValueError: No headers based on this input can be generated. Please relax or change some of the requirements you specified.
```

## Root cause

`generate_headers()` pins the version exactly:

```python
browsers = [Browser(name="chrome", min_version=ver, max_version=ver)]
```

with `ver = 149`, but the installed dataset stops earlier. Measured on `browserforge==1.2.4` /
`apify-fingerprint-datapoints==0.14.0`:

| browser | newest shipped |
|---|---|
| chrome | 143 |
| firefox | 144 |
| edge | 141 |

Since `generate_headers()` runs at module scope in both `fingerprints.py` and `_config_tools.py`,
this is an import-time crash rather than a degraded fingerprint. The call at `fingerprints.py:59`
survives because it also offers firefox/edge and browserforge treats the list as alternatives; the
two chrome-only calls in `_config_tools.py` are the ones that raise.

## Why not just lower the constant

Lowering `chrome_version` to 143 works today and breaks again on the next bump — that has happened at
every release since 0.4.2 (145 → 147 → 148 → 149). The constants are meant to track the browser
being driven, so they should be free to move ahead of the dataset. The data is also no longer
something a user can refresh: as of browserforge 1.2.4 it ships in the separate
`apify-fingerprint-datapoints` package and `python -m browserforge update` is a deprecated no-op. So
the version gap can only be closed by a release of that package, and Scrapling shouldn't be unusable
in the meantime.

This PR leaves the constants at 149 and makes the mismatch non-fatal instead.

## The fix

Keep the exact pin as the happy path, and add an `except ValueError` fallback that picks the newest
version the installed dataset can actually be used with:

- **Candidates come from the datapack**, so the search is bounded by real data rather than a guessed
  range.
- **Each candidate is confirmed through `HeaderGenerator` before it's used.** Being shipped isn't
  sufficient on its own — the OS has to line up too. This is the part that matters: **on Linux the
  newest shipped Chrome (143) still fails, and only 141 is usable.** Selecting the newest shipped
  version would leave #396 (the Linux report) unfixed.
- Both lookups are `lru_cache`d, so the probe runs at most once per (browser, ceiling, OS).

Worst case is 5 `generate()` calls; the happy path adds none.

## Before / after

Requesting the pinned 149:

| OS | before | after |
|---|---|---|
| macOS | `ValueError` | `Chrome/143` |
| Linux | `ValueError` | `Chrome/141` |
| Windows | `ValueError` | `Chrome/143` |

The happy path is unchanged — with a version the dataset supports, the UA reports exactly that
version and the fallback never executes (asserted in the tests):

| requested | result | fallback ran |
|---|---|---|
| 143 (macOS) | `Chrome/143` | no |
| 130 | `Chrome/130` | no |
| 999 | `Chrome/143` | yes |

## Tests

Added to `TestFingerprintFunctions` in `tests/fetchers/test_utils.py`:

1. the browser fetchers import and expose non-empty default user agents;
2. headers are still generated with the version monkeypatched to `999`, across all three
   `browser_mode` values;
3. a supported version is used *exactly*, with an assertion that the fallback did **not** run — so a
   future change can't silently downgrade a working case.

Test 3 discovers the supported version at runtime instead of hardcoding it, so it won't rot as the
datapack moves.

Full suite locally: **814 passed**, with 4 failures / 2 collection errors that are present on a clean
`dev` checkout too and unrelated to this change (all `mcp.server.fastmcp` under mcp 2.0.0, i.e. #392).
Before this change the same suite could not even collect: 40 collection errors and 411 passed.

## Relationship to the other open PRs

There are three open PRs for this bug, which I found only after writing this one — happy to close
this if you prefer one of them, but the platform differences are worth recording:

| | approach | macOS | Linux | Windows |
|---|---|---|---|---|
| #397 | drop the version pin on fallback | `Chrome/140` | `Chrome/141` | `Chrome/141` |
| #398 | use the newest version in the datapack | `Chrome/143` | **`ValueError`** | `Chrome/143` |
| this | newest datapack version *confirmed usable* | `Chrome/143` | `Chrome/141` | `Chrome/143` |

- **#398** has the right idea and this PR is essentially it plus one check. As written it still
  crashes on Linux, because it assumes a version present in the datapack can be generated for.
- **#397** does stop the crash everywhere, but dropping `min_version` lets browserforge sample the
  whole range — it returns `Chrome/140` on macOS where 143 is available, which weakens the pinning the
  comment above the constants describes as deliberate.
- **#395** is complementary rather than competing: it defers the module-scope calls to session
  startup. That's worth doing on its own merits — it's what turned this data gap into a total import
  failure for everyone, including users who only wanted the parser — but it relocates this particular
  error rather than removing it, since `generate_headers(browser_mode=True)` still raises when the
  session actually starts. #395 and this change fit together.

### DIFF
--- a/scrapling/engines/toolbelt/fingerprints.py
+++ b/scrapling/engines/toolbelt/fingerprints.py
@@ -5,8 +5,10 @@ Functions related to generating headers and fingerprints generally
 from functools import lru_cache
 from platform import system as platform_system
 
+from apify_fingerprint_datapoints import get_browser_helper_file
 from browserforge.headers import Browser, HeaderGenerator
 from browserforge.headers.generator import SUPPORTED_OPERATING_SYSTEMS
+from orjson import loads
 
 from scrapling.core._types import Dict, Literal, Tuple
 
@@ -34,6 +36,47 @@ def get_os_name() -> OSName | Tuple:
             return SUPPORTED_OPERATING_SYSTEMS
 
 
+@lru_cache(4, typed=True)
+def _dataset_versions(name: str) -> Tuple:
+    """Get the major versions of a browser that the installed fingerprints dataset ships with.
+
+    :param name: The browser name as browserforge knows it
+    :return: The versions found, ordered from the newest to the oldest
+    """
+    prefix = f"{name}/"
+    entries = loads(get_browser_helper_file().read_bytes())
+    versions = {int(entry[len(prefix) :].split(".")[0]) for entry in entries if entry.startswith(prefix)}
+    return tuple(sorted(versions, reverse=True))
+
+
+@lru_cache(8, typed=True)
+def _newest_supported_version(name: str, ceiling: int, os_name: OSName | Tuple) -> int:
+    """Find the newest version of a browser that the installed browserforge dataset can still generate headers for.
+
+    :param name: The browser name as browserforge knows it
+    :param ceiling: The wanted version, no version newer than it is considered
+    :param os_name: The same OS constraint the caller uses, as it affects which versions are usable
+    :return: The newest usable version, or `ceiling` itself if none was found
+    """
+    # Shipping a version doesn't mean headers can be generated for it, as the OS and the device
+    # have to line up with it as well, so each candidate is confirmed before we settle on it
+    for version in _dataset_versions(name):
+        if version > ceiling:
+            continue
+
+        try:
+            HeaderGenerator(
+                browser=[Browser(name=name, min_version=version, max_version=version)],
+                os=os_name,
+                device="desktop",
+            ).generate()
+            return version
+        except ValueError:
+            continue
+
+    return ceiling
+
+
 def generate_headers(browser_mode: bool | str = False) -> Dict:
     """Generate real browser-like headers using browserforge's generator
 
@@ -53,7 +96,15 @@ def generate_headers(browser_mode: bool | str = False) -> Dict:
                 Browser(name="edge", min_version=140),
             ]
         )
-    return HeaderGenerator(browser=browsers, os=os_name, device="desktop").generate()
+    try:
+        return HeaderGenerator(browser=browsers, os=os_name, device="desktop").generate()
+    except ValueError:
+        # The versions above are pinned to the browsers we drive, but browserforge's dataset is shipped
+        # separately and can lag behind them. Degrading to the newest version it does know about keeps the
+        # pinning intent while stopping that data gap from turning into an import-time crash.
+        supported = _newest_supported_version("chrome", ver, os_name)
+        browsers[0] = Browser(name="chrome", min_version=supported, max_version=supported)
+        return HeaderGenerator(browser=browsers, os=os_name, device="desktop").generate()
 
 
 __default_useragent__ = generate_headers(browser_mode=False).get("User-Agent")
--- a/tests/fetchers/test_utils.py
+++ b/tests/fetchers/test_utils.py
@@ -1,3 +1,5 @@
+import re
+
 import pytest
 
 from scrapling.engines.toolbelt.convertor import ResponseFactory
@@ -8,6 +10,7 @@ from scrapling.engines.toolbelt.navigation import (
     create_async_intercept_handler,
     _is_domain_blocked,
 )
+from scrapling.engines.toolbelt import fingerprints
 from scrapling.engines.toolbelt.fingerprints import get_os_name, generate_headers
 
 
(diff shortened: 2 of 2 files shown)


D4Vinci:
Nice idea, buddy, but it's not the optimal solution. I will make scrapling access Apify's neural network right away to skip the generation part, so it's faster, and also I want it now to be not limited to desktop devices only.

This whole issue was caused by an [issue](https://github.com/apify/fingerprint-suite/issues/564) with `apify-fingerprint-datapoints` version `0.14.0`, but they have now released version `0.15.0`, which fixed it. Anyway, I will make sure this doesn't happen again

I will wait for your future PRs; don't let this discourage you from making more PRs in the future!

PR #410, opened 2026-08-12
fix(parser): find/find_all with class_ silently miss multi-class elements

yetval:
Hey! If an element has more than one class, looking it up by one of them comes back empty. No error, just nothing:

```python
page = Selector(content='<div class="product featured">ONLY PRODUCT</div>')
page.css('.product')                    # finds it  ✓
page.find_all('div', class_='product')  # []  ← empty, even though it's right there
```

It looks for a class of exactly `"product"`, but classes are a list of words, so anything with a second class slips through. `css()` and `has_class` both find it, and the README shows them as the same thing.

Tiny fix so it matches `css()` and BeautifulSoup. Added tests in `TestFindByClass` that fail on `dev` today, rest still pass.

Written by me, reviewed by Claude Code.


D4Vinci:
Nice catch @yetval as usual 😄

Issue #413, opened 2026-08-13
Require authentication by default for HTTP transport


Paupu:
### What would you like to share?

When the MCP server is started with `--http` (Streamable HTTP transport), it accepts unauthenticated requests by default. A `--auth-token` flag (or `SCRAPLING_MCP_AUTH_TOKEN` env var) can be passed to require bearer token authentication, but without it the server starts with a warning:

```
The MCP server is running over HTTP without authentication, so anyone who can reach 0.0.0.0:8000 can use every tool, including fetching arbitrary URLs from this machine.
Pass `--auth-token` (or set the SCRAPLING_MCP_AUTH_TOKEN environment variable) to require a bearer token.
```

The server starts anyway.

HTTP mode exposes the server on a network interface. Anyone who can reach the port can invoke every tool (for example fetch arbitrary URLs, open browser sessions, take screenshots). A warning log line is not strong enough as a security control. New users who start the server with `--http` without realizing auth is optional will have an unprotected scraping server on their network.

Make authentication **required** for HTTP mode by default. Require users to explicitly opt out if they want unauthenticated access. This will also 

### New flag: `--no-auth` (or `--allow-unauthenticated`)

```bash
# Default: auth required
scrapling-mcp --http

# Opt out: allow unauthenticated access (not recommended)
scrapling-mcp --http --no-auth
```

### Behavior change

| Mode | Current | After |
| --- | --- | --- |
| `scrapling-mcp --http` | Starts, logs warning, no auth required | Starts, auth required (401 without token) |
| `scrapling-mcp --http --auth-token xxx` | Starts, auth required | Same (explicit token) |
| `scrapling-mcp --http --no-auth` | N/A | Starts, no auth required (explicit opt-out) |
| `scrapling-mcp --stdio` | No auth (stdio is local-only) | No change |

### Implementation notes

- In `serve()`, check `if http and not self._auth_token and not self._allow_unauthenticated:`
  - If true, either require `--auth-token` or `--no-auth` at startup (fail with error)
  - If neither is provided, default to requiring auth (i.e. treat as if `--auth-token` was set)
- The `--no-auth` flag sets a flag that skips the auth middleware
- `--auth-token` continues to work as before (sets the token to verify against)
- `--stdio` is unaffected (stdio is local-only, no auth needed)

### Backward compatibility

[...]

PR #414, opened 2026-08-17
fix: require authentication by default for the HTTP transport

yamantaka-singh:
Hey!

This is a go at #413 — making the MCP server's Streamable HTTP transport actually require authentication instead of just warning that there isn't any.

I saw the issue is marked as something you'd keep in mind rather than something you'd asked anyone to build, so no hard feelings at all if the timing's off or you'd rather solve it a different way. Happy to rework or close it.

## The problem

Today `scrapling-mcp --http` logs a warning and then serves anyway:

```
The MCP server is running over HTTP without authentication, so anyone who can reach
0.0.0.0:8000 can use every tool, including fetching arbitrary URLs from this machine.
```

The warning is accurate, but it scrolls past in a second and the server is already up. Since the default host is `0.0.0.0`, someone trying `--http` for the first time gets an open scraping server on their network without really deciding to.

## What this changes

`--http` on its own now stops and tells you what to pick:

```console
$ scrapling-mcp --http
Usage: scrapling mcp [OPTIONS]
Try 'scrapling mcp --help' for help.

Error: Refusing to serve the MCP server over HTTP without authentication because anyone
who can reach 0.0.0.0:8000 would be able to use every tool, including fetching arbitrary
URLs from this machine. Pass `--auth-token` (or set the SCRAPLING_MCP_AUTH_TOKEN
environment variable) to require a bearer token, or `--no-auth` to serve it
unauthenticated anyway.
```

And there's a new `--no-auth` flag for people who genuinely want the old behaviour:

```bash
scrapling-mcp --http --no-auth   # still warns, but serves
```

| Command | Before | After |
| --- | --- | --- |
| `scrapling-mcp --http` | serves, warns, no auth | refuses to start, exit code 2 |
| `scrapling-mcp --http --auth-token xxx` | serves, auth required | unchanged |
| `scrapling-mcp --http --no-auth` | n/a | serves, warns, no auth |
| `scrapling-mcp` (stdio) | no auth | unchanged |

One small extra: passing `--auth-token` *and* `--no-auth` together keeps the token and warns, rather than quietly throwing your token away. Silently dropping a credential because of flag ordering felt like the wrong way to lose an argument.

## Where the guard lives

I put the check in `ScraplingMCPServer.serve()` rather than in `cli.py`, so anyone using the server as a library gets it too:

```python
server = ScraplingMCPServer()
server.serve(http=True, host="0.0.0.0", port=8000)   # raises ValueError
server.serve(http=True, host="0.0.0.0", port=8000, allow_unauthenticated=True)  # fine
```

`cli.py` just catches that `ValueError` and re-raises it as a click `UsageError`, so CLI users get a clean error message instead of a traceback.

`stdio` is deliberately untouched — it's only reachable by the process that spawned it, so there's nothing to authenticate against.

## Backward compatibility

This is a breaking change for anyone running `--http` with no token, which I know isn't nothing. `--no-auth` is a one-word fix for them, and the error message names the flag directly so nobody has to go hunting through docs to get unstuck. Your call entirely on whether that trade is worth it — if you'd rather ship it behind a deprecation cycle first, I'm glad to redo it that way.

## Testing

Added tests in `tests/ai/test_ai_mcp.py` and `tests/cli/test_cli.py` covering the refusal, the `--no-auth` opt-out, the token-wins-over-opt-out rule, and stdio staying unaffected.

I also ran the real server and checked it over the wire rather than trusting the mocks:

| Server | Request | Status |
| --- | --- | --- |
| `--http --no-auth` | no token | reaches the MCP layer (no auth gate) |
| `--http --auth-token abc123` | no token | `401` |
| `--http --auth-token abc123` | wrong token | `401` |
| `--http --auth-token abc123` | correct token | reaches the MCP layer |

Local run of the non-browser suite: 876 passed. `ruff`, `bandit`, `vermin`, `mypy`, and `pyright` are all clean on the changed files. (`ruff` does report 50 pre-existing issues elsewhere in `tests/`, but those are identical on `dev` — I left them alone rather than mixing an unrelated cleanup into this PR.)

Also updated the authentication sections of `docs/ai/mcp-server.md` and `agent-skill/Scrapling-Skill/references/mcp-server.md` to match the new default.

---

### AI disclosure

Per [AI_POLICY.md](https://github.com/D4Vinci/Scrapling/blob/dev/AI_POLICY.md): this PR was written with Claude Code — it explored the codebase, wrote the implementation, the tests, and this description. I reviewed everything, ran the tests and the live-server checks above myself, and I understand and stand behind the change. Happy to explain any part of it in my own words.


tahazarif10:
One thing I'd change with this is the default bind address

The docs now suggest `--http --no-auth` for local testing but the default host is still `0.0.0.0` so that command actually exposes the server to the LAN

Defaulting HTTP to `127.0.0.1` and requiring an explicit `--host 0.0.0.0` for network access would make the secure default much harder to bypass accidentally

yamantaka-singh:
Good catch, you're right and it's a hole I opened myself. Pushed a fix in b2cc38e.

I checked it before changing anything, since I wanted to be sure it wasn't just theoretical. Running the exact command my docs recommended:

```
$ scrapling-mcp --http --no-auth --port 8941
$ lsof -nP -iTCP:8941 -sTCP:LISTEN
Python  61593 ... TCP *:8941 (LISTEN)

$ curl -o /dev/null -w '%{http_code}\n' http://10.199.67.120:8941/mcp
400
```

`*:8941` and a real response on the LAN address, so an unauthenticated scraping server was genuinely reachable from the network. Same command after the change:

```
$ lsof -nP -iTCP:8942 -sTCP:LISTEN
Python  63497 ... TCP 127.0.0.1:8942 (LISTEN)

$ curl --max-time 5 http://10.199.67.120:8942/mcp
(connection refused)
```

So `--host` now defaults to `127.0.0.1`, and `--host 0.0.0.0` is the explicit opt-in for network access. That splits it into two separate decisions instead of one: "can the network reach this" and "does it need a token". Docs updated to match, plus a line warning against pairing `--no-auth` with `--host 0.0.0.0`, and a test pinning both the default and the override.

Worth flagging the cost, since it's a second breaking change on top of the auth one: anyone running the container with `-p 8000:8000 ... mcp --http` now needs `--host 0.0.0.0` or the port publish silently does nothing. I went ahead because the documented Docker setup in the README and `server.json` both use stdio, so I don't think this breaks a documented path — but @D4Vinci, if you'd rather keep `0.0.0.0` and only fix the docs, say the word and I'll drop the commit.

*(Disclosure per AI_POLICY.md: this reply and the fix were written with Claude Code. I ran the checks above and reviewed the diff.)*

D4Vinci:
Thanks @yamantaka-singh for solving this. It's missing some stuff, but that's fine; I will add them.

PR #417, opened 2026-08-19
fix(parser): keep filtering on blank class_ and escape CSS string values

yamantaka-singh:
Hey! Two small follow-ups to #410, both in the same spot.

The first is an edge case in the new `class_` handling. If the class value is blank, the filter quietly disappears:

```python
page = Selector('<div class="">empty</div><div class="a">a</div><div>none</div>')
page.find_all('div', class_='')   # all three ← filter is gone
```

`"".split()` is `[]`, so the join adds nothing, the selector stays `*`, and the `if selector != "*"` check drops it — then `find_all` takes its "no filters given" path and hands back everything. Before #410 it built `*[class=""]` and returned the one element. Not something you'd type on purpose, but `class_=some_var` with an empty var silently returns more than you asked for.

The second is older. A line break in an attribute value raises instead of matching:

```python
page = Selector('<a title="one\ntwo">link</a>')
page.find_all('a', title='one\ntwo')   # SelectorSyntaxError
```

CSS strings can't hold a raw line break, so it goes in as a hex escape. Quotes were already handled, newlines weren't.

Both fixed in the same loop: a blank class falls through to exact match, and values go through a small escape helper.

One I couldn't fix — backslashes in values still don't match. cssselect loses characters while unescaping them (`\\` and `\5C ` both eat the backslash and the character after it), so nothing we emit can match. Left it alone with a note in the docstring rather than guessing at a workaround.

Tests in `TestFindByClass` plus a new `TestFindByEscapedAttributes` — both fail on `dev` today, rest of the suite passes.

Written with Claude Code; I ran the checks and reviewed the diff myself.


D4Vinci:
LGTM, thanks for catching this one!

PR #418, opened 2026-08-19
fix(ai): keep session settings on MCP session fetches

Yigtwxx:
## Proposed change

The MCP tools that take a `session_id` (`fetch`, `bulk_fetch`, `stealthy_fetch`, `bulk_stealthy_fetch`, `screenshot`) forward **every** per-fetch parameter to the session's `fetch()`, including their own defaults when the caller supplied nothing:

```python
entry.session.fetch(
    url,
    wait=wait,                              # 0
    timeout=timeout,                        # 30000
    google_search=google_search,            # True
    ...
    solve_cloudflare=solve_cloudflare,      # False
)
```

`validate_fetch` (`scrapling/engines/_browsers/_validators.py`) treats every key it receives as an override, so the branch that reads the value from the session config is skipped for all of them:

```python
for key in fetch_param_fields:
    if key in kwargs_dict:
        overrides[key] = kwargs_dict[key]
    elif hasattr(session, "_config") and hasattr(session._config, key):
        result[key] = getattr(session._config, key)     # never reached for those keys
```

The tool defaults happen to be identical to the `PlaywrightConfig` defaults, so `_filter_defaults` strips them all and `validate(overrides, model)` returns an all-defaults config, which then overwrites the session's values. The comment above that code — *"This prevents validated defaults from overwriting session config values"* — is defeated precisely because the caller passes the defaults explicitly.

The effect is that the settings a session was opened with are dropped on every fetch made through it. `solve_cloudflare` is the sharpest case, because `validate_fetch` deliberately preserves it whenever it appears in the overrides:

```python
from scrapling.engines._browsers._validators import validate_fetch, StealthConfig

session = AsyncStealthySession(solve_cloudflare=True)
print(session._config.solve_cloudflare, session._config.timeout)
# True 60000

# what `bulk_stealthy_fetch` sends today when the caller passes only `urls` and `session_id`
params = validate_fetch(
    dict(
        wait=0, timeout=30000, google_search=True, extra_headers=None,
        disable_resources=False, wait_selector=None, wait_selector_state="attached",
        network_idle=False, solve_cloudflare=False,
    ),
    session,
    StealthConfig,
)
print(params.solve_cloudflare, params.timeout)
# False 30000
```

So a stealthy session opened with `solve_cloudflare=True` never reaches `if params.solve_cloudflare:` in `AsyncStealthySession.fetch`, and it also loses the 60s timeout that `StealthConfig.__post_init__` sets for it. The same happens to `wait`, `timeout`, `google_search`, `extra_headers`, `disable_resources`, `wait_selector`, `wait_selector_state` and `network_idle` on both dynamic and stealthy sessions.

The session-less paths were already correct — they call `session.fetch(url)` with no keyword arguments — so this only affects the MCP layer, not the library API.

### The change

Both paths now pass only the parameters the caller actually supplied, through a small helper next to the existing `_page_pool_size`:

```python
def _supplied_params(**params: Any) -> Dict[str, Any]:
    """Keep only the parameters the caller set, so the unset ones fall back to the session/fetcher defaults."""
    return {name: value for name, value in params.items() if value is not None}
```

Those parameters now default to `None` on the five tools. `proxy` is left as it was: both `fetch` implementations pop it before validation, so passing `None` and omitting it are already the same thing.

Behaviour without a session is unchanged. Every affected parameter has a tool default identical to its `PlaywrightConfig`/`StealthConfig` default, and `validate` filters out defaults before converting, so omitting a key and passing its default take the same code path. The `test_a_fresh_session_is_built_with_the_same_settings_as_before` test asserts that by building a real session from the kwargs the tools produce and checking the resulting config.

I also dropped `proxy` from the "ignored when using a session" list in the server-level instructions, since it is honoured per request, and noted that unset options now defer to the session.

### Tests

Added to `tests/ai/test_ai_mcp.py`, none of which launch a browser or touch the network:

- `TestSessionSettingsSurviveFetchKwargs` builds real sessions and runs the kwargs the tools now send through `validate_fetch`, asserting the session's own settings survive — including `solve_cloudflare` and its 60s timeout — and that supplied values still override the session.
- `TestSessionParameterForwarding` uses the existing fake-session pattern to assert what actually reaches `session.fetch()`, and that a fresh session is still built with the same effective settings as before.

Five of the new tests fail on `dev` and pass with the change. `_FakeAsyncBrowserSession` now records its fetch kwargs and runs a `page_action` if it gets one, which is what lets the screenshot tool be covered without a browser.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

One consequence worth flagging: with `None` as the "not supplied" marker, a caller can no longer pass `wait_selector=null` or `extra_headers=null` to clear a value the session was opened with. That only worked as a side effect of the bug, and it seemed the wrong trade against having every unset option silently reset the session.

Verified locally on Python 3.12: the full `tests/ai/test_ai_mcp.py` suite passes (69 tests), and `ruff check`, `ruff format --check`, `bandit`, `vermin`, `mypy` and `pyright` are all clean.

### Checklist:

### AI assistance disclosure

Per [AI_POLICY.md](https://github.com/D4Vinci/Scrapling/blob/main/AI_POLICY.md): I used Claude Code while working on this. It helped me trace the parameter flow from the MCP tools into `validate_fetch`, and it wrote a first pass of the patch and the tests. I confirmed the defect myself by running the repro above against `dev`, checked each affected parameter's tool default against its struct default before relying on the omission being a no-op, reviewed the diff and the test assertions, and ran the suite and the quality checks locally.


D4Vinci:
LGTM, thanks for catching this one!

D4Vinci:
Hi, just a quick update.

The changes introduced issues with normal fetches without sessions, since they use the same tools as sessions. Also, this PR prevents the AI agent from being aware of the default values for all parameters. So, I removed almost all the changes in this PR and restructured the MCP server to eliminate the bug class you pointed out, while keeping the AI agent aware of the default values for all parameters.

You can see the commit above if you are interested.

Yigtwxx:
Thanks for taking the time to explain it — both objections land, and both were mine to catch.

On the first one: those tools also serve session-less callers, and the table I put in the body proved the tool defaults *equalled* the struct defaults. That answers "do the values match?" and not "how many kinds of caller does this function have?", which was the question that mattered. On the second: on an MCP surface the signature is the documentation the model reads, so `None` defaults bought the supplied/not-supplied distinction by spending every visible default value. Splitting `open_session` (browser-level) from `session_fetch` (per-request) removes the ambiguity instead of encoding it, and returning the session's effective `settings` shows the agent what I was trying to infer on its behalf. Better fix than mine.

I did read `73ad18b`, and one thing there looks like a regression rather than a design choice:

`session_fetch` accepts and documents `proxy`, but a per-request proxy goes through `_page_generator` in `scrapling/engines/_browsers/_base.py`, which needs `self.browser` — and `self.browser` is only set when the session was opened with `cdp_url` or a `proxy_rotator`. A plain session runs on `launch_persistent_context`, so it stays `None`:

```python
import asyncio
from scrapling.fetchers import AsyncDynamicSession


async def main():
    async with AsyncDynamicSession(headless=True) as session:
        print("browser attr =", session.browser)  # None
        await session.fetch("http://127.0.0.1:1/", proxy="http://127.0.0.1:9999")


asyncio.run(main())
# RuntimeError: Browser not initialized for proxy rotation mode
```

It raises before any navigation, so no network or live site is involved. Since `proxy` was also dropped from `open_session`, MCP sessions now have no working proxy path on either side, while the session-less `fetch`/`stealthy_fetch` still proxy fine through the constructor.

Happy to open a PR if you want one, but I would rather not guess the direction here: restore a session-level `proxy` on `open_session`, or make the per-request path work without a rotator? Your call and I will follow it.

---

Disclosure: this comment was written with AI assistance (Claude), per `AI_POLICY.md`. The reproduction above was run locally against `dev` at `73ad18b`.

D4Vinci:
Thanks @Yigtwxx for pointing that out; I forgot that we are using persistent context by default, but that was going to change anyway since it's now the default to use one tab per session on the MCP server, so it's meaningless to set the proxy per tab. I have moved it to be a session-level argument in the commit above.

Yigtwxx:
That is a cleaner framing than either option I offered — with one tab per session there is no per-tab scope left for a proxy to mean anything in.

Checked `93ae0ac` against `dev` locally, and a session-level proxy now reaches the persistent context:

```
config.proxy:          {'server': 'http://127.0.0.1:9999', ...}
context_options proxy: {'server': 'http://127.0.0.1:9999', ...}
fetch -> net::ERR_PROXY_CONNECTION_FAILED
```

It tries the proxy now instead of raising before navigation. Thanks for the credit in the commit message.

---

Disclosure: this comment was written with AI assistance (Claude), per `AI_POLICY.md`. The check above was run locally against `dev` at `0cb3e97`.

PR #419, opened 2026-08-19
fix(spiders): keep the request meta on cached responses

Yigtwxx:
## Proposed change

`ResponseCacheManager.get` rebuilds a `Response` from the cached JSON without passing `meta`, so `Response.__init__` sets `self.meta = {}`. The cache-hit path in `CrawlerEngine` reattaches the request but never merges its meta back in:

```python
cached = await self._cache_manager.get(request._fp)
if cached is not None:
    cached.request = request
    ...
    await self._run_callbacks(request, cached)
```

The live path does merge it, in `SessionManager.fetch`:

```python
response.request = request
# Merge request meta into response meta (response meta takes priority)
response.meta = {**request.meta, **response.meta}
```

So the first run of a `development_mode` spider behaves correctly and the second one hands every callback an empty `response.meta`. Anything the spider carried between requests is silently gone: `response.follow(..., meta=...)` chains lose their state, and `ShopifySpider` raises

```
KeyError: 'handle'
```

on `response.meta["handle"]` (`scrapling/spiders/templates/shopify.py`) as soon as its responses start coming from the cache — which is exactly what `development_mode` is for, iterating on parsing code without refetching.

The fix mirrors the live path on the cache-hit path, with the response's own meta still winning:

```python
cached.request = request
# Cached responses are rebuilt without meta, so merge the request's in as the live path does
cached.meta = {**request.meta, **cached.meta}
```

### Tests

`test_cached_run_keeps_the_request_meta` in `tests/spiders/test_cache.py` reuses the existing `MockSpider`/`MockSession` harness: it starts a request with `meta={"page": 7}`, runs the engine twice over the same cache directory, and asserts the callback sees the same meta on the cached run as on the live one. It fails on `dev` with `[{'meta': {}}]` and passes with this change.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

This is the same shape as #379, where the cache dropped browser-engine cookies; meta is the other field the rebuilt response never gets back.

Verified locally on Python 3.12: `tests/spiders/` passes, and `ruff check`, `ruff format --check`, `bandit`, `vermin`, `mypy` and `pyright` are clean.

### Checklist:

### AI assistance disclosure

Per [AI_POLICY.md](https://github.com/D4Vinci/Scrapling/blob/main/AI_POLICY.md): I used Claude Code while working on this. It helped me find the asymmetry between the cache-hit path and `SessionManager.fetch`, and drafted the one-line fix and the test. I reproduced the empty meta myself against `dev`, traced it back to `ResponseCacheManager.get` not passing `meta` to `Response`, confirmed `ShopifySpider` is affected, and ran the suite and the quality checks locally.


D4Vinci:
Thanks @Yigtwxx for catching this one!

PR #427, opened 2026-08-27
fix: stop retrying libcurl errors that a retry cannot change

### LINKED ISSUE #426: Static fetcher retries certificate verification failures three times

### Python version

Python 3.13.14

### Scrapling version

0.4.15. Also reproduced on current main at `458e2a2ac909b3235747ebcdb312b93a1080a10a`.

### Dependencies version

Installed with uv. Relevant packages:

```text
click==8.5.0
curl-cffi==0.16.2
scrapling==0.4.15
```

macOS 15.7.7, arm64

Yes. Scrapling is installed as a uv tool.

### Expected behavior

Certificate verification failures such as an expired certificate should fail after the first attempt. Retrying the same request cannot change the certificate validation result.

### Actual behavior

Scrapling retries the request three times before returning the same certificate error. The current main request loop catches every `CurlError` and retries it without classifying deterministic TLS verification failures.

### Steps To Reproduce

```sh
scrapling extract get "https://expired.badssl.com/" /tmp/tls.md
```

Observed:

```text
WARNING: Attempt 1 failed: ... curl: (60) ... certificate has expired ... Retrying in 1 seconds...
WARNING: Attempt 2 failed: ... curl: (60) ... certificate has expired ... Retrying in 1 seconds...
ERROR: Failed after 3 attempts: ... curl: (60) ... certificate has expired ...
```

The command exits 1 after repeating the deterministic failure. The same behavior occurs on current main.

This issue was investigated and drafted with Codex. I reproduced it on 0.4.15 and current main and reviewed the static request retry path.

alkaz-nodemaven:
## Proposed change

`_make_request` catches every `CurlError` and retries it. Some of those errors describe the client's own configuration, or the URL it was handed, and repeating the identical request cannot turn them into a success - it only adds `retry_delay` to the time before the caller learns what went wrong. #426 reports this for an expired certificate.

This adds a classification step in front of the existing retry, in both the sync and the async loop. It does not change the retry machinery itself.

**The classification is deliberately not a flat blacklist, because one group of codes depends on the rotator:**

- `_FATAL_CURL_CODES` - the request is wrong regardless of the route: `UNSUPPORTED_PROTOCOL` (1), `URL_MALFORMAT` (3), `BAD_FUNCTION_ARGUMENT` (43), `SSL_ENGINE_NOTFOUND` (53), `SSL_ENGINE_SETFAILED` (54), `SSL_CERTPROBLEM` (58), `SSL_CIPHER` (59), `SSL_CACERT_BADFILE` (77). The three SSL codes in that list are about the local client's own certificate, key, CA file or cipher list, not about anything the peer sent. Never retried.
- `_PEER_CERTIFICATE_CURL_CODES` - `PEER_FAILED_VERIFICATION` (60) and `SSL_ISSUER_ERROR` (83). These are deterministic for a given peer, which is #426's case, but the loop already calls `self._proxy_rotator.get_proxy()` at the top of every attempt, so with a rotator the next attempt is a different exit. A TLS-intercepting middlebox on one exit is a real cause of a code 60 that the next exit does not reproduce. So these fail fast only when the next attempt would be identical - no rotator, or a per-request `proxy=` that has pinned the route.

Everything else keeps its current behaviour. `OPERATION_TIMEDOUT` and `COULDNT_RESOLVE_HOST` are still retried; the second one showed up on its own during the runs below, on a DNS blip, and was correctly retried away.

Two smaller decisions worth stating, since both could reasonably have gone the other way:

- Matching is on `CurlError.code`, not on the message text. The codes are stable across libcurl versions and the wording is not.
- I did not extend the existing `is_proxy_error` for this. That helper matches lowercase substrings because it also has to classify Playwright errors, which carry no libcurl code; here the code is available and is the better key. The two are used for different things - `is_proxy_error` only picks the wording of a log line.

### Measurements

The reporter's own command, on this machine, Windows 10, Python 3.11.0, curl_cffi 0.16.2, no proxy, against `dev` at `458e2a2`.

Before:

```
$ scrapling extract get "https://expired.badssl.com/" /tmp/tls.md
WARNING: Attempt 1 failed: ... curl: (60) ... certificate has expired ... Retrying in 1 seconds...
WARNING: Attempt 2 failed: ... curl: (60) ... certificate has expired ... Retrying in 1 seconds...
ERROR: Failed after 3 attempts: ... curl: (60) ... certificate has expired ...
```

After:

```
$ scrapling extract get "https://expired.badssl.com/" /tmp/tls.md
ERROR: Request failed with an error that a retry cannot change: ... curl: (60) ... certificate has expired ...
```

Exit code is 1 either way and the same `CertificateVerifyError` reaches the caller; only the two pointless attempts and the two `retry_delay` sleeps are gone. Through the library rather than the CLI, with `FetcherSession(retries=3, retry_delay=1)`, counting the retry log lines: 3 attempts in 2.98 s and 3.72 s before, 1 attempt in 0.42 s and 1.09 s after, two runs a side.

### What I ran

- `pytest tests/fetchers` - **38 failed, 197 passed** with the patch and **38 failed, 189 passed** without it. The `FAILED` lists are identical, so the 8 new tests are the whole difference. Those 38 are this machine's missing extras: `markdownify` from `[rag]`, and the Playwright browser binaries. I skipped `test_pages.py`, `test_cloudflare_solver.py` and `test_page_reuse.py` for the same reason, so **I have not reproduced a fully green suite locally** and CI is the real check.
- `ruff check` and `ruff format --diff` at the pinned 0.14.5 - clean.
- `bandit -r -c .bandit.yml` - 0 issues.
- `vermin -t=3.10- --violations --eval-annotations` - passes.
- `mypy scrapling/` - `Success: no issues found in 58 source files`.
- `pyright scrapling/` - 12 errors, none in a file this PR touches, all of them imports of optional extras that CI installs with `.[all]` and I did not.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

- This PR fixes or closes an issue: fixes #426

### AI assistance disclosure

Per [AI_POLICY.md](https://github.com/D4Vinci/Scrapling/blob/main/AI_POLICY.md): this PR was written with Claude Code. It wrote the patch, the tests and this description. I directed the work, chose the rotator-dependent split over a flat blacklist after reading the loop, and ran every command quoted above myself on my own machine. I understand the change and can defend any line of it. If I answer a review comment with AI help I will say so in that comment.

### Checklist:

### DIFF
--- a/scrapling/engines/static.py
+++ b/scrapling/engines/static.py
@@ -4,7 +4,7 @@ from time import sleep as time_sleep
 from asyncio import sleep as asyncio_sleep
 
 from curl_cffi.curl import CurlError
-from curl_cffi import CurlHttpVersion
+from curl_cffi import CurlECode, CurlHttpVersion
 from curl_cffi.requests import (
     BrowserTypeLiteral,
     Session as CurlSession,
@@ -31,6 +31,47 @@ from ._browsers._types import RequestsSession, GetRequestParams, DataRequestPara
 from .toolbelt.fingerprints import generate_headers, __default_useragent__
 
 _NO_SESSION: Any = object()
+
+# libcurl codes that describe the client's own configuration or the URL it was
+# given. Sending the identical request again cannot turn any of them into a
+# success, so a retry only adds `retry_delay` to the time before the caller
+# learns what went wrong.
+_FATAL_CURL_CODES = frozenset(
+    {
+        CurlECode.UNSUPPORTED_PROTOCOL,
+        CurlECode.URL_MALFORMAT,
+        CurlECode.BAD_FUNCTION_ARGUMENT,
+        CurlECode.SSL_ENGINE_NOTFOUND,
+        CurlECode.SSL_ENGINE_SETFAILED,
+        CurlECode.SSL_CERTPROBLEM,
+        CurlECode.SSL_CIPHER,
+        CurlECode.SSL_CACERT_BADFILE,
+    }
+)
+
+# Verification failures for the certificate the peer presented. Deterministic
+# for that peer, but a rotator hands out a different exit on the next attempt,
+# and a TLS-intercepting middlebox on one route is a real cause of these - so
+# they are only worth failing fast on when the request would be identical.
+_PEER_CERTIFICATE_CURL_CODES = frozenset(
+    {
+        CurlECode.PEER_FAILED_VERIFICATION,
+        CurlECode.SSL_ISSUER_ERROR,
+    }
+)
+
+
+def _is_retryable(error: CurlError, rotating: bool) -> bool:
+    """Whether sending the same request again could plausibly return something else.
+
+    `rotating` says whether a proxy rotator will pick a different exit for the
+    next attempt, which is what decides the peer-certificate cases.
+    """
+    if error.code in _FATAL_CURL_CODES:
+        return False
+    if error.code in _PEER_CERTIFICATE_CURL_CODES:
+        return rotating
+    return True
 
 
 def _select_random_browser(impersonate: ImpersonateType) -> Optional[BrowserTypeLiteral]:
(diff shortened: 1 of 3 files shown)


D4Vinci:
We leave it like that intentionally, so it's up to the user. Too much hand-holding is bad; most of our users are developers.
Regarding the ticket you mentioned, you can control this behavior with the --no-verify flag, which disables certificate verification so the error won't happen.

PR #430, opened 2026-08-31
fix: keep the stealth flags when the user passes extra_flags

alkaz-nodemaven:
## Proposed change

`extra_flags` is documented as "a list of **additional** browser flags to pass to the
browser on launch" - in `chrome.py`, `stealth_chrome.py`, `docs/fetching/dynamic.md` and
`docs/fetching/stealthy.md`. On a `StealthySession` it does not add flags, it replaces
them: passing one flag of your own removes all 53 `STEALTH_ARGS` and every conditional
hardening flag with them.

`__generate_stealth_options` builds the whole hardening set and hands it to
`__generate_options__` as the `extra_flags` **argument**:

```python
# _base.py, StealthySessionMixin.__generate_stealth_options
flags = tuple(DEFAULT_ARGS) + tuple(STEALTH_ARGS)
if config.block_webrtc: flags += (...)
if not config.allow_webgl: flags += (...)
if config.hide_canvas: flags += (...)
super(StealthySessionMixin, self).__generate_options__(flags)
```

and `__generate_options__` picks one of the two lists with `or`:

```python
# _base.py:482
flags = list(set(tuple(flags) + tuple(config.extra_flags or extra_flags or ())))
#                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

`config.extra_flags or extra_flags` returns the user's list when there is one, so the
bundle is discarded. Nothing warns; the session still starts and still calls itself
stealthy.

Measured on `dev` @ `458e2a2`, reading `_browser_options["args"]` - the dict
`chromium.launch(**self._browser_options)` is called with - for
`StealthySession(headless=True, block_webrtc=True, allow_webgl=False, hide_canvas=True)`:

| | flags | `STEALTH_ARGS` kept | `DEFAULT_ARGS` kept |
|---|---|---|---|
| no `extra_flags` | 70 | 53/53 | 11/11 |
| `extra_flags=["--window-size=1280,720"]` | **12** | **0/53** | 11/11 |

The test flag is deliberately one that is in neither list, so "the user's flag arrived"
and "the bundle arrived" are independent observations.

Individually, in the second case:

| flag | without `extra_flags` | with |
|---|---|---|
| `--webrtc-ip-handling-policy=disable_non_proxied_udp` | present | **gone** |
| `--disable-webgl` | present | **gone** |
| `--fingerprinting-canvas-image-data-noise` | present | **gone** |
| `--disable-blink-features=AutomationControlled` | present | **gone** |

Same thing on the real Chrome command line, read back off the running process through
`Win32_Process` rather than off the options object, `extra_flags` arm:

```
dev 458e2a2 :  45 switches   webrtc ABSENT   webgl ABSENT   canvas noise ABSENT
patched     : 103 switches   webrtc YES      webgl YES      canvas noise YES
```

One caveat on that table, because it would be easy to overclaim: Playwright emits
`--disable-blink-features=AutomationControlled` itself, so on the command line that one
survives either way. It is lost from Scrapling's list, not from the browser. The other
three are genuinely gone.

### The `set()` in the same line

`set()` over strings is ordered by a per-process hash seed, so the flag list differed
from run to run. Five subprocesses, `dev`:

```
distinct orderings : 5/5      (same set of flags: True)
--disable-features landed at index 22 in one run, 28 in another, 48, 59, ...
```

Chrome parses `--disable-features` and `--lang` **last-wins** - given the switch twice it
keeps the last value and silently drops the earlier one whole - so which one wins was
decided by the hash seed. Today `STEALTH_ARGS` declares only one `--disable-features`, so
there is nothing visible to lose yet; the point is that it is a coin flip waiting for a
second one, and that a launch that is not reproducible run to run is hard to debug.

`dict.fromkeys` de-duplicates the same way and keeps first-seen order, so this is a
one-word change in the same line as the fix, not a separate concern.

After the patch, same probe:

```
no extra_flags              : 70 flags, 53/53 STEALTH_ARGS
extra_flags=[one new flag]  : 71 flags, 53/53 STEALTH_ARGS, user's flag present, last
distinct orderings          : 1/5
```

User flags are concatenated **last**, so where a switch is last-wins the user's value
still wins - which is what somebody reaching for `extra_flags` is asking for.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

The diff is one statement in `_base.py` plus the tests that reproduce it, which
`CONTRIBUTING.md` asks for on a bugfix. The `set()` -> `dict.fromkeys` change is in that
same statement; splitting it out would mean touching the line twice.

No documentation change: the docs already describe the behaviour this restores.

### Additional information

I could not find an existing issue for this. It is not the sort of thing that gets
reported, because the failure is silent - the session starts, fetches, and returns a
page. It only shows up as a detection rate that got worse after somebody added a flag.

**AI disclosure, per [AI_POLICY.md](https://github.com/D4Vinci/Scrapling/blob/main/AI_POLICY.md).**
I used Claude Code on this one. It wrote the patch, the three tests and this PR body. The
`or` was something I went looking for after seeing a stealth session come up with far
fewer flags than expected; the probe scripts, the base-vs-patched runs and the
command-line readback were directed and run by me, and the numbers above are from those
runs on my own machine. I understand what the change does and can defend it. If I answer
a review comment here with AI help, I will say so on that comment.

One thing worth reporting from the end-to-end experience: my first version of the control
test asserted that `DynamicSession` "behaved correctly before this fix". It did not - it
never lost a flag, because it passes no bundle for `or` to discard, but it went through
the same `set()` and had its order shuffled too. The test caught my own wrong claim when
I ran it against unpatched `dev`. The test now covers both halves separately, so a future
change that fixes one and reintroduces the other is caught.

### Checklist:
* [x] All new Python files are placed inside an existing directory. -- no new files; the tests go in `tests/fetchers/sync/test_stealth_session.py`

---

**How this was tested**

- `pytest tests/fetchers tests/parser tests/cli` - **19 failed / 380 passed on `dev`,
  19 failed / 383 passed with this branch.** The failing sets are byte-identical
  (diffed, not counted): 15 in `tests/cli` and 4 in `test_response_markdown.py`, all
  pre-existing on this host and unrelated. The delta is exactly the 3 new tests.
- The 3 new tests were run against unpatched `dev` first and **all three fail** there.
  A regression test that passes before the fix is not testing the fix.
- `ruff check scrapling/` and `ruff format --diff` clean; `mypy` clean on the changed file.
- Base and patched numbers were produced in the same session on the same host, so the
  two sides differ only in the patch.


scrapling/engines/_browsers/_base.py
```diff
@@ -479,7 +479,15 @@ def __generate_options__(self, extra_flags: Tuple | None = None) -> None:
...
-                flags = list(set(tuple(flags) + tuple(config.extra_flags or extra_flags or ())))
+                # `extra_flags` is the caller's bundle - for a stealth session it is the
+                # whole hardening set - and `config.extra_flags` is the user's own. Both
+                # have to be kept; `a or b` used to pick one and silently drop the other,
+                # so a user passing a single flag lost every stealth flag with it.
+                # The user's flags go last, so they win the switches Chrome parses
+                # last-wins (`--disable-features`, `--lang`, ...).
+                # Order is preserved on purpose: `set()` here made the flag list differ
+                # from run to run, which decides those last-wins switches by coin flip.
+                flags = list(dict.fromkeys(tuple(flags) + tuple(extra_flags or ()) + tuple(config.extra_flags or ())))
```
D4Vinci: Use `set` instead of `dict.fromkeys` and remove all of these comments.

D4Vinci (changes requested):

D4Vinci:
Thanks, @alkaz-nodemaven, for catching and fixing this!

PR #431, opened 2026-08-31
fix(ai): match inline style values instead of substrings when sanitizing

Yigtwxx:
## Proposed change

The prompt-injection sanitizer decides what is hidden with XPath `contains()` on the whole `style` attribute:

```python
# scrapling/core/shell.py
_HIDDEN_XPATH = XPath(
    './/*[contains(@style,"display:none") or contains(@style,"display: none")'
    ...
    ' or contains(@style,"opacity:0") or contains(@style,"opacity: 0")'
    ' or contains(@style,"font-size:0") or contains(@style,"font-size: 0")'
    ' or contains(@style,"height:0") or contains(@style,"height: 0")'
    ' or contains(@style,"width:0") or contains(@style,"width: 0")]'
    ...
)
```

`contains()` is a plain substring test, so a value that merely *starts* with a zero matches, and so does a different property whose name *ends* with a matching one:

| inline style | substring that matches | what it actually means |
|---|---|---|
| `opacity:0.95` | `opacity:0` | almost fully opaque |
| `font-size:0.9rem` | `font-size:0` | slightly smaller text |
| `width:0.5em` | `width:0` | half an em wide |
| `line-height:0.9` | `height:0` | a different property entirely |
| `min-height:0` | `height:0` | the flexbox default |

`_sanitize_for_ai` then calls `drop_tree()` on the match, so the element **and its whole subtree** are deleted.

This was reachable only through the opt-in MCP path when it was added in `375951b`. v0.4.15 changed that: `Response.markdown()` runs the sanitizer unconditionally, and `SiteToMarkdownSpider` puts an entire site through it.

```python
from scrapling.engines.toolbelt.custom import Response

html = ('<html><body><p>Intro.</p>'
        '<div style="font-size:0.9rem"><h1>Chapter 1</h1><p>The whole article body.</p></div>'
        '</body></html>')
response = Response(url="https://example.com/", content=html, status=200, reason="OK",
                    cookies={}, headers={}, request_headers={})

response.markdown()
# 'Intro.'
```

`Chapter 1` and the article body are gone, with no warning. `line-height:0.9`, `opacity:0.95` and `width:0.5em` behave the same, and with `main_content_only=True` on a page whose `<body>`-level wrapper carries one of those styles, `markdown()` returns `''`.

The MCP server hits it too, on its default `main_content_only=True`:

```python
from scrapling.core.shell import Convertor

"".join(Convertor._extract_content(response, "markdown", None, True))
# 'Intro.'
```

### The change

The XPath now only selects candidates, and a small helper confirms them by comparing declarations instead of substrings:

```python
_HIDDEN_XPATH = XPath(".//*[@style] | .//*[@aria-hidden='true'] | .//template")
_HIDING_DECLARATIONS = frozenset({("display", "none"), ("visibility", "hidden")})
_ZERO_HIDING_PROPERTIES = frozenset({"opacity", "font-size", "height", "width", "max-height", "max-width"})
_ZERO_VALUE_PATTERN = re_compile(r"0(?:\.0+)?[a-z%]*")
```

`_is_hidden_element` splits the attribute on `;`, splits each declaration on the first `:`, and requires the property to be one of the hiding ones *and* its value to actually be zero. `0`, `0px`, `0%` and `0.0em` still count; `0.9`, `0.95rem` and `0.5em` no longer do. `!important` is stripped off the value, so `display:none !important` keeps working, and both spacings (`display:none`, `display: none`) fall out of the parsing instead of needing their own XPath arm.

`drop_tree()` is now guarded by that check, which is the only change inside `_sanitize_for_ai`.

The property set is chosen to keep every genuine hiding case the old XPath caught: `display:none`, `visibility:hidden`, `opacity:0`, `font-size:0`, `height:0`, `width:0`, plus `max-height:0` and `max-width:0`, which the old `contains(@style,"height:0")` arm also caught. The only true positives it drops are `min-height:0` and `line-height:0` — neither hides anything.

### Tests

Two tests in `tests/fetchers/test_response_markdown.py`, both browser-free:

- `test_partial_zero_values_are_not_treated_as_hidden` covers `line-height:0.9`, `opacity:0.95`, `font-size:0.9rem`, `width:0.5em` and `min-height:0`. On `dev` it fails with `assert 'Compact heading' in ''` — the whole body is dropped.
- `test_truly_hidden_content_is_still_removed` pins the defence: `display:none`, `display: none`, `visibility:hidden`, `opacity:0`, `opacity: 0`, `height:0px`, `max-height:0`, `width:0`, `font-size:0`, `display:none !important`, `aria-hidden="true"` and `<template>` all stay out. It passes before and after, so it is there to prove the fix does not narrow the sanitizer.

The existing `test_sanitization_is_always_applied` is untouched and still passes. `pytest tests` gives 927 passed on Python 3.12; the only failures are four pre-existing Windows-only ones in `tests/spiders/test_checkpoint.py` and `tests/spiders/test_result.py` that fail the same way on an unmodified `dev`. `ruff check`, `ruff format --check`, `mypy`, `pyright`, `bandit -c .bandit.yml` and `vermin -t=3.10-` are all clean locally.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

Two scope decisions worth flagging, both easy to change if you would rather go the other way:

- I kept `max-height` and `max-width` in the zero set purely to preserve today's behaviour, since `contains(@style,"height:0")` used to catch them. If you consider a bare `max-height:0` too weak a hiding signal without `overflow:hidden`, dropping them is a one-word edit.
- I left the docs alone. `docs/ai/mcp-server.md` describes the sanitizer by the categories it removes, not by how it matches them, so nothing there becomes wrong with this change.

### Checklist:

### AI assistance disclosure

Per [AI_POLICY.md](https://github.com/D4Vinci/Scrapling/blob/main/AI_POLICY.md): I used Claude Code while working on this. It helped me trace the sanitizer from `_HIDDEN_XPATH` through `_sanitize_for_ai` to both callers, and it drafted the helper and the tests. I found and reproduced the defect myself, chose the property set so the change stays behaviour-preserving, checked that no existing test or doc pinned the substring matching, reviewed the diff, and ran the suite and the quality checks locally.


D4Vinci (changes requested):
Thanks, this fixes a real issue with visible content being removed.

Before merging, please handle CSS comments in style values. For example, `display:none/*hidden*/` hides the element in the browser, but the new helper does not detect it.

Please add a regression test for that case and `opacity:0/*hidden*/`, while keeping the tests that preserve visible decimal values.

Yigtwxx:
Handled in 727e143. `_is_hidden_element` now strips `/* ... */` before splitting the declarations, so `display:none/*hidden*/` and `opacity:0/*hidden*/` are dropped again, matching what the browser does and what the old substring XPath happened to catch.

Regression cases added for both, plus a comment inside a declaration (`display:/* keep */none`) and a comment containing a separator (`color:red/*;*/;visibility:hidden`). `opacity:0.95/*hidden*/` went into the visible set so the decimal cases stay covered.

The branch is also rebased on the current `dev`: the `.//slot[@hidden]` candidate from 6bbc6e0 is part of the rewritten XPath and confirmed by the helper, so `test_shadow_hidden_slots.py` passes unchanged.

Local run: ruff, bandit, vermin, mypy and pyright are clean; the affected test directories pass (263 tests).

AI assistance disclosure: the follow-up patch, its tests and this comment were drafted with Claude Code. I reviewed the diff, checked the browser behaviour the review describes, and ran the checks and tests locally.

D4Vinci:
Thanks @Yigtwxx, the comment handling and hidden-slot support look good. There’s one remaining case: `display:none ! important` and `opacity:0 ! important` are valid CSS. Chrome hides both, but the helper misses them because it only strips the exact `!important` suffix.

Please handle whitespace between `!` and `important` and add regression tests.

Yigtwxx:
Handled in c94a3c9. The `!important` suffix is now removed with `!\s*important$` instead of an exact `removesuffix`, so `display:none ! important` and `opacity:0 ! important` are compared as `none` and `0` again.

Regression cases added for both, plus `visibility:hidden!important` for the no-space form. `opacity:0.95 ! important` went into the visible set so the suffix handling does not reopen the decimal case.

Local run: ruff, bandit, vermin, mypy and pyright are clean; `tests/fetchers`, `tests/parser` and `tests/core` pass (460 tests).

AI assistance disclosure: the patch, its tests and this comment were drafted with Claude Code. I reviewed the diff and ran the checks and tests locally.

D4Vinci:
Thanks, the `! important` fix looks good. I found a few more cases where Chrome hides the element, but the new helper keeps its text:

- `opacity:00`
- `opacity:0e0`
- `display:none/*unfinished`
- `--a:"/*";display:none;--b:"*/"`

The current sanitizer removes all four. The zero regex misses valid zero forms, and the comment regex treats quoted comment markers as real comments while missing comments that run to the end of the attribute.

Please add regression tests for these cases and adjust the matching before we merge.

Yigtwxx:
Handled in 6c08f8a. The two regexes are gone: `_split_style_declarations` walks the style once, dropping comments (including one that runs to the end of the attribute) and copying quoted strings and unquoted `url(...)` tokens as they are, so a `;` or `/*` inside them neither ends a declaration nor opens a comment. The zero check now matches a CSS `<number>` token and compares it as a float, so `00`, `0e0`, `.0`, `+0` and `0.` all count.

Regression cases added for the four you listed plus `opacity:.0` and `url(/*);display:none;url(*/)`. The visible set gained `opacity:5e-1`, `--note:"a;display:none;b"` and `url(a;display:none;b)`; the quoted one was a false positive with the old split, since the `;` inside the string ended the declaration.

Not covered on purpose: identifier escapes (`display:n\6fne`). The old XPath never matched them either, so it is not a regression; I can add it if you want it in.

The branch is already on top of `dev`. Local gates (ruff, bandit, vermin, mypy, pyright) and the full suite pass.

AI assistance disclosure: the tokenizer and the test cases were drafted with Claude; I reproduced each case against the helper, reviewed the diff and ran the suite locally.

D4Vinci:
Thanks for the work on this. The original issue is valid, but the latest revision introduces further problems:
- Removing comments joins separate CSS words. For example, Chrome ignores `display:no/**/ne`, but the helper turns it into `display:none` and removes visible content.
- The number regex becomes very slow on long invalid values. A value containing 6,000 zeros followed by `!` took about 0.8 seconds in local testing.

I’m closing this PR because the implementation keeps growing while introducing new faults.

**Please open a new PR** with a focused fix that adds no dependencies, fixes the original false matches, and preserves the existing genuine hidden-content cases, including hidden slots, templates, and `aria-hidden`.

Please include regression tests covering the cases discussed throughout this PR and before/after benchmarks. The goal is reliable inline-style matching within the sanitizer’s existing scope, not a full CSS parser.

PR #432, opened 2026-08-31
fix(response): keep css_selector matches as separate markdown blocks

Yigtwxx:
## Proposed change

`Response.markdown()` joins the converted matches with an empty string:

```python
# scrapling/engines/toolbelt/custom.py
pages = [page] if not css_selector else page.css(css_selector)
return "".join(Convertor._convert_to_markdown(element.html_content) for element in pages)
```

`markdownify` returns each chunk with its trailing newlines already stripped, so joining with `""` welds the last line of one match onto the first line of the next. That is not only a cosmetic run-on: when the next match starts with a heading, the setext underline that follows ends up applying to the merged line, so the previous paragraph is absorbed into the heading.

```python
from scrapling.engines.toolbelt.custom import Response

html = ('<html><body>'
        '<article><h2>First section</h2><p>Body one.</p></article>'
        '<article><h2>Second section</h2><p>Body two.</p></article>'
        '</body></html>')
response = Response(url="https://example.com/", content=html, status=200, reason="OK",
                    cookies={}, headers={}, request_headers={})

response.markdown(css_selector="article")
# 'First section\n-------------\n\nBody one.Second section\n--------------\n\nBody two.'
```

`Body one.Second section` is now a single H2, so `Body one.` has been promoted out of the body text and into a heading. With plain blocks it is simply glued: `markdown(css_selector=".a")` on two `<div class="a">` gives `'First paragraph.Second paragraph.'`.

The sibling path does not have this problem. `Convertor._extract_content`, which the MCP server uses, yields the chunks separately, so the same "all matches are concatenated" contract behaves differently depending on which entry point you go through.

It also hits the flow the RAG docs recommend — *"Use `css_selector` on the spider to cut boilerplate before chunking"* — because `SiteToMarkdownSpider` passes its `css_selector` straight into `markdown()`, so a whole crawled corpus inherits the welded blocks.

### The change

One line: join with a blank line so each match stays its own Markdown block.

```python
return "\n\n".join(Convertor._convert_to_markdown(element.html_content) for element in pages)
```

A blank line is the block separator Markdown already uses between the elements *inside* a single match, so the output is consistent whether two paragraphs come from one match or two. The single-match case is unaffected, since there is nothing to join.

### Tests

In `tests/fetchers/test_response_markdown.py`:

- `test_css_selector_concatenates_all_matches` already existed but only asserted `"First" in md` and `"Second" in md`, which passes both before and after the fix. It now asserts `"First\n\nSecond"`, so it actually pins the separation.
- `test_css_selector_matches_do_not_bleed_into_each_other` is new and covers the heading case, asserting both that `"Body one.\n\nSecond section"` is present and that `"Body one.Second section"` is not.

Both fail on `dev` and pass with the change. `_make_response` grew a `content` parameter that defaults to the existing `HTML` constant, so every other test in the file is untouched.

`pytest tests` gives 982 passed on Python 3.12; the only failures are four pre-existing Windows-only ones in `tests/spiders/test_checkpoint.py` and `tests/spiders/test_result.py` that fail the same way on an unmodified `dev`. `ruff check`, `ruff format --check`, `mypy`, `pyright`, `bandit -c .bandit.yml` and `vermin -t=3.10-` are all clean locally.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

I kept the change to `Response.markdown()` on purpose, and would rather ask than guess about the other one: `Convertor.write_content_to_file` does the same `"".join` over `_extract_content`, so `scrapling extract get <url> out.md --css-selector "article"` writes the same welded blocks to the file. I did not touch it here because `_extract_content` ends with an unconditional `yield ""` outside its loop, so switching that join to `"\n\n"` would also leave a trailing separator at the end of every file, and that is a separate decision about the generator rather than about this one-line fix. Happy to fold it into this PR or open a second one, whichever you prefer.

The docs need no change either way: both the docstring and `docs/ai/building-rag-systems.md` say the matches are concatenated, which stays true.

### Checklist:

### AI assistance disclosure

Per [AI_POLICY.md](https://github.com/D4Vinci/Scrapling/blob/main/AI_POLICY.md): I used Claude Code while working on this. It helped me compare `Response.markdown()` against `Convertor._extract_content` and drafted the tests. I found and reproduced the defect myself, confirmed the existing test could not catch it, decided to keep `write_content_to_file` out of scope rather than change the generator, reviewed the diff, and ran the suite and the quality checks locally.


D4Vinci:
Thanks @Yigtwxx for catching this one.

PR #435, opened 2026-09-03
fix(fetchers): fail fast on deterministic TLS certificate errors

### LINKED ISSUE #426: Static fetcher retries certificate verification failures three times

AnandkumarMall:
Resolves #426. Adds an \is_ssl_verification_error\ check to prevent the static and async fetchers from repeatedly retrying when hitting deterministic TLS issues like expired certificates (curl error code 60). Includes tests verifying that SSL errors raise immediately while transient errors still retry.

### DIFF
--- a/scrapling/engines/static.py
+++ b/scrapling/engines/static.py
@@ -26,7 +26,7 @@ from scrapling.core._types import (
 
 from .toolbelt.custom import Response
 from .toolbelt.convertor import ResponseFactory
-from .toolbelt.proxy_rotation import ProxyRotator, is_proxy_error
+from .toolbelt.proxy_rotation import ProxyRotator, is_proxy_error, is_ssl_verification_error
 from ._browsers._types import RequestsSession, GetRequestParams, DataRequestParams, ImpersonateType
 from .toolbelt.fingerprints import generate_headers, __default_useragent__
 
@@ -258,10 +258,13 @@ class _SyncSessionLogic(_ConfigurationLogic):
                     assert response is not None
                     result = ResponseFactory.from_http_request(response, selector_config, meta={"proxy": proxy})
                     return result
-                except CurlError as e:  # pragma: no cover
+                except CurlError as e:
                     if attempt < max_retries - 1:
-                        # Now if the rotator is enabled, we will try again with the new proxy
-                        # If it's not enabled, then we will try again with the same proxy
+                        # TLS certificate failures are deterministic: the same cert
+                        # will produce the same error on every attempt. Fail fast.
+                        if is_ssl_verification_error(e):
+                            log.error(f"TLS certificate verification failed (not retrying): {e}")
+                            raise
                         if is_proxy_error(e):
                             log.warning(
                                 f"Proxy '{proxy}' failed (attempt {attempt + 1}) | Retrying in {retry_delay} seconds..."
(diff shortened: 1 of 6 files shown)


Diogo-Damasceno (commented):
Verified locally (Arch Linux, Python 3.14.7, Scrapling from source at `28c3296`).

I checked the part that matters most here: that transient errors still retry. Exercising `is_ssl_verification_error` against representative curl messages:

```
curl: (60) SSL certificate problem: certificate has expired  -> True
certificate verify failed: self signed certificate           -> True
curl: (28) Operation timed out                               -> False
curl: (7) Failed to connect to host                          -> False
```

So timeouts and connection failures are not caught by the new branch and keep their existing retry behaviour. The new tests (6) pass with the change.

On `main` without it, `tests/fetchers/test_ssl_cert_error.py` cannot even be collected (`ImportError: cannot import name 'is_ssl_verification_error'`), so the tests do depend on the new helper.

Note on my environment, not on the PR: `tests/fetchers` + `tests/core` show 18 collection errors both with and without this change (missing optional modules here), so they are pre-existing. `tests/core` alone passes 43/43 in both cases.

One wording suggestion, non-blocking: `ssl: certificate_verify_failed` in `_SSL_CERT_ERROR_INDICATORS` looks like a Python/urllib3 spelling rather than something curl emits, but keeping it is harmless and makes the check cover non-curl backends too.

AnandkumarMall:
Friendly ping! Let me know if there's anything else needed from my side or if this is ready for review. Thanks!

Diogo-Damasceno:
Hey @AnandkumarMall, nice work — I tested this locally and the logic checks out: SSL fails fast and transient errors still retry as expected.

A few notes to make the PR cleaner:

- `uv.lock` got included in the diff (1808 lines) but the repo doesn't track a lockfile; it looks like it came from a local `uv` run. Probably best to drop it from the commit.
- the `except` lines have `# pragma: no cover`, but that branch is now covered by the tests. Safe to remove the pragma there.
- not a maintainer myself, but worth noting: the CI checks are stuck on "waiting" because fork workflows need D4Vinci's approval to run. That's why no status shows up.

Rest looks good.

AnandkumarMall:
Thanks for the review @Diogo-Damasceno! Updated:
- Dropped \uv.lock\ from the PR diff
- Removed \# pragma: no cover\ from the \except CurlError as e:\ lines now that they are covered by tests
- Rebased on \upstream/main\ to bring the branch up to date

D4Vinci:
Thanks, but we leave it like that intentionally, so it's up to the user.
As stated in the issue ticket, you can control this behavior with the `--no-verify` flag, which disables certificate verification and prevents the error.

PR #438, opened 2026-09-05
fix: keep robots.txt policies separate for HTTP and HTTPS

maximilliangrand:
## Proposed change

When `robots_txt_obey=True`, HTTP and HTTPS URLs sharing a hostname reuse the same cached robots.txt parser. For example, after checking an HTTP URL whose policy allows `/private/`, an HTTPS URL can also be allowed even when its own robots.txt disallows that path. Reversing the fetch order can wrongly block allowed pages. The engine also reuses the first scheme's crawl delay and prefetches only one scheme from `start_urls`.

Cache parsers by their robots.txt URL, cache resolved delays by scheme and authority, and prefetch each distinct origin. Requests to different paths on the same origin still share their cached policy and delay. The getting-started guide now explains that HTTP and HTTPS policies are separate.

This follows the scope of robots.txt described in [RFC 9309 section 2.3](https://www.rfc-editor.org/rfc/rfc9309.html#section-2.3), also illustrated explicitly in [Google's protocol/host/port examples](https://developers.google.com/crawling/docs/robots-txt/robots-txt-spec#file-location).

### Reproduction and validation

Six regression cases exercise independent HTTP/HTTPS allow/disallow rules and delay directives, both lookup orders, concurrent prefetching, and duplicate seed paths. The engine delay tests isolate its own cache from the parser cache, so fixing just the parser cannot make them pass. All six reproduced the defect on unmodified `dev`.

On Python 3.13, the full suite passes using the project's separate async, synchronous, and browser groups:

- `pytest tests/ -m asyncio -k 'not (DynamicFetcher or StealthyFetcher)' --cov=scrapling`: 291 passed.
- `pytest tests/ -m 'not asyncio' -k 'not (DynamicFetcher or StealthyFetcher)' -n 4 --cov=scrapling --cov-append`: 675 passed.
- `pytest tests/ -k 'DynamicFetcher or StealthyFetcher' --cov=scrapling --cov-append`: 25 passed.
- Combined coverage: 91%.
- Final focused run after extending the delay regression to different paths: 111 passed.
- `pre-commit run --all-files`, `mypy scrapling/`, and `pyright --pythonpath .venv/bin/python scrapling/`: passed.

### AI assistance disclosure

Investigation, implementation, regression tests, documentation, and this PR description were primarily written with Codex. The change was checked against the existing cache and scheduling paths and validated using the tests above.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

- This PR fixes or closes an issue: fixes # (no existing issue; reproduction included above)
- This PR is related to an issue: # (none)
- Link to documentation pull request: **Included in this PR.**

### Checklist:

### DIFF
--- a/docs/spiders/getting-started.md
+++ b/docs/spiders/getting-started.md
@@ -174,11 +174,11 @@ class PoliteSpider(Spider):
 
 When enabled, the spider will:
 
-1. **Pre-fetch robots.txt** for all domains in `start_urls` before the crawl begins (concurrently).
-2. **Check every request** against the domain's robots.txt `Disallow` rules. Disallowed requests are silently dropped and counted in `stats.robots_disallowed_count`.
+1. **Pre-fetch robots.txt** for all origins in `start_urls` before the crawl begins (concurrently).
+2. **Check every request** against the origin's robots.txt `Disallow` rules. Disallowed requests are silently dropped and counted in `stats.robots_disallowed_count`.
 3. **Respect `Crawl-delay` and `Request-rate` directives** by taking the maximum of the directive and your configured `download_delay`. This means robots.txt delays never reduce your configured delay, only increase it when needed.
 
-Robots.txt files are fetched using the spider's default session and cached per domain for the entire crawl. Domains discovered mid-crawl (not in `start_urls`) have their robots.txt fetched on the first request to that domain.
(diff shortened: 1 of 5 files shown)


D4Vinci:
Thanks for your contribution, but I will close this as an unrealistic edge case. I have never seen a website that makes different rules for HTTP and HTTPS. I don't think anyone would ever need to do that, and I couldn't find a website with two different robots.txt files to support this idea.

PR #444, opened 2026-09-11
fix(cli): gracefully skip playwright install-deps on non-apt Linux

### LINKED ISSUE #439: `scrapling install` unconditionally runs apt-based Playwright dependency installation on Arch Linux

- [x] I searched open and closed issues for `arch`, `apt-get`, and `install-deps`, and PRs for `apt-get`. The returned Arch reports (#103, #118) concern browser runtime behavior; #57 concerns spaces in Python paths, not this dependency-installation failure.

3.12.14

0.4.15; source checkout at 28c329671485daaea89a40fb34a7db8622e51468 (main).

### Dependencies version

Relevant installed dependency: playwright==1.61.0. This is a focused dependency-installer report, not a full environment freeze.

Arch Linux family, specifically Omarchy 4.0.2, x86_64 (`ID=omarchy`, `ID_LIKE=arch`). `apt-get` is not installed or available on PATH.

Yes, a local Python 3.12 virtual environment.

### Expected behavior

On non-apt Linux distributions, `scrapling install` should allow Chromium download without unconditionally attempting apt-based system dependency installation. It should explain that users must install any missing Chromium libraries with their distribution package manager. This need not imply official Playwright support for Arch Linux or guarantee that the browser can launch without the required libraries.

### Actual behavior and verified evidence

The upstream `install()` implementation always runs `python -m playwright install-deps chromium` after downloading Chromium. `__Execute()` uses `subprocess.check_output()` without catching the subprocess failure, so a dependency-installation failure aborts the command before the TLD update and installation marker.

Source: https://github.com/D4Vinci/Scrapling/blob/28c329671485daaea89a40fb34a7db8622e51468/scrapling/cli.py

A safe probe of the actual installed Playwright dependency installer on this machine:

```console
$ .venv/bin/python -m playwright install-deps --dry-run chromium
Failed to install browser dependencies
Error: Failed to run 'apt-get install -s' to simulate dependency install: spawn apt-get ENOENT
```

Exit status: 1. stderr also reports:

```text
BEWARE: your OS is not officially supported by Playwright; installing dependencies for ubuntu24.04-x64 as a fallback.
```

The dry run confirms that this environment falls back to an apt-based installer even though apt-get does not exist. I did not run a full unpatched `scrapling install --force` during this verification, to avoid downloads and privileged installation attempts; the Scrapling failure path above is based on inspection of the upstream code.

### Steps To Reproduce

1. Use Arch Linux or an Arch-based distribution without `apt-get`.
2. Install Scrapling with browser-fetcher dependencies in a virtual environment.
3. Run `python -m playwright install-deps --dry-run chromium` to observe the missing-apt failure without installing packages.
4. The upstream `scrapling install --force` path invokes the same dependency installer without the dry-run flag, after downloading Chromium.

### Proposed direction / local workaround

A local, uncommitted patch guards the automatic system-dependency step with a Linux/`apt-get` availability check, retains Chromium download, and prints distribution-package-manager guidance when skipping on non-apt Linux. It includes two focused CLI tests and documentation updates. The patch restored cleanly after pulling current main, but those tests have not been run against the updated checkout, and browser launch is not claimed to be verified.

The intended change is graceful handling of unsupported automatic dependency installation, not a blanket suppression of installation failures on supported systems.

AnandkumarMall:
Resolves #439.

### Problem
On Linux distributions that do not use `apt-get` (such as Arch Linux, Manjaro, Fedora, Alpine, Void, NixOS, etc.), running `scrapling install` crashes with:
```text
Error: Failed to run 'apt-get install -s' to simulate dependency install: spawn apt-get ENOENT
subprocess.CalledProcessError: Command '['.../python', '-m', 'playwright', 'install-deps', 'chromium']' returned non-zero exit status 1.
```
This halts installation prematurely, preventing TLD updates and leaving `.scrapling_dependencies_installed` uncreated.

### Solution
1. In `scrapling.cli.install`: Check if running on Linux and whether `shutil.which("apt-get")` is available before invoking `playwright install-deps chromium`.
2. When `apt-get` is absent on Linux, print a message informing the user that system dependencies can be installed using their distribution package manager if Chromium needs them, and continue with the setup.
3. Added CLI unit tests covering:
   - Linux without `apt-get` (skips `install-deps`, prints guidance, completes setup)
   - Linux with `apt-get` / supported systems (executes `install-deps`)
   - Re-running `install` when already installed
4. Added documentation note in `docs/cli/overview.md`.

### DIFF
--- a/docs/cli/overview.md
+++ b/docs/cli/overview.md
@@ -28,3 +28,6 @@ and the installation of the fetchers' dependencies with the following command
 scrapling install
 ```
 This downloads all browsers, along with their system dependencies and fingerprint manipulation dependencies.
+
+!!! note "Non-Debian Linux distributions (Arch Linux, Fedora, Alpine, etc.)"
+    On Linux systems where `apt-get` is unavailable, `scrapling install` downloads the Chromium browser and completes the setup, but skips automatic system package installation. If Chromium fails to launch, install any missing system libraries using your distribution's package manager.
--- a/scrapling/cli.py
+++ b/scrapling/cli.py
@@ -1,7 +1,8 @@
 from os import environ
 from pathlib import Path
+from shutil import which as shutil_which
 from subprocess import check_output
-from sys import executable as python_executable
+from sys import executable as python_executable, platform as sys_platform
 
 from scrapling import __version__
 from scrapling.core.utils import log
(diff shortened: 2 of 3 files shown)


D4Vinci:
Thanks for the contribution, but we are going to close this one as explained in the issue ticket.

PR #446, opened 2026-09-14
fix(spiders): cache responses only after blocked-response checks

Kuang-xianxin:
## Proposed change

With `development_mode=True`, a response is cached before `is_blocked` runs. A retry of that URL then hits the cached blocked page and sends it straight to the parse callback. For example, a session returning a blocked response followed by a successful response is fetched only once, and the callback receives the blocked body.

Write responses to the development cache only after they pass the spider's live `is_blocked` hook. Cache replay remains unchanged: reconstructed responses may lack metadata needed by custom hooks. Users with blocked pages cached by older versions can clear their development cache once.

Three regressions cover a blocked HTTP 403, blocked content with HTTP 200, and exhausted retries. The retry cases also run a second crawl using a custom hook that requires live-only metadata, verifying that the accepted response is replayed without a fetch or another call to that hook. Tests exercise the real engine, scheduler, requests, and disk cache with a controlled session.

Issue investigation, code, tests, and this description were prepared with AI assistance.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

- This PR fixes or closes an issue: none; reproduced locally on `dev` at `6e64404`.
- This PR is related to an issue: none.
- Link to documentation pull request: not applicable.

Validation on Python 3.13.14 / Windows, after merging current `dev` (`2beadcf`):

- All three regressions fail against unmodified `dev`, then pass with the final fix. The live-only metadata cases also fail on the previous PR implementation, reproducing the review concern.
- `pytest tests/spiders/test_cache.py tests/spiders/test_engine.py tests/spiders/test_request.py tests/spiders/test_sitemap.py tests/parser tests/core -q`: **273 passed**.
- `pre-commit run --all-files`: all four hooks pass. Mypy passes for all 58 source files; Pyright reports zero errors using the project's virtualenv.
- The full repository and browser suites were not rerun for this follow-up. The earlier full-suite attempt hit documented pre-existing Windows checkpoint/path and file-cleanup failures; those are not counted as a successful full-suite run.

### Checklist:


scrapling/spiders/engine.py
```diff
@@ -202,13 +202,14 @@ async def _process_request(self, request: Request) -> None:
...
-                await self._run_callbacks(request, cached)
-                return
+                if not await self.spider.is_blocked(cached):
+                    self.stats.cache_hits += 1
+                    self.stats.increment_requests_count(request.sid or self.session_manager.default_session_id)
+                    self.stats.increment_response_bytes(request.domain, len(cached.body))
+                    self.stats.increment_status(cached.status)
+                    log.debug(f"Cache hit: {request.url}")
+                    await self._run_callbacks(request, cached)
+                    return
```
D4Vinci, line 212: Don't change this part; leave it as it is.

D4Vinci (changes requested):
Thanks, please keep the change that caches responses only after they pass `is_blocked()`, but leave cache replay unchanged. Cached responses do not preserve all live response metadata, so calling custom `is_blocked()` hooks on them can cause errors.

Users can clear old development caches once. Please adjust the tests to cover this smaller change.

scrapling/spiders/engine.py
Kuang-xianxin: Addressed in 194145b. The cache-hit branch is restored unchanged; the production diff now only moves cache writes after the live blocked-response check. I removed the old-cache revalidation test and kept three regression cases. The retry cases use a hook requiring live-only metadata and verify that the second crawl replays the accepted response without calling that hook.

Validation: the three cases fail on unmodified current dev (2beadcf), while the previous PR implementation fails both metadata-dependent replay cases. The final related suite passes 273 tests; full pre-commit, mypy and Pyright pass. The PR description now notes that old development caches can be cleared once.

Code, tests, and this reply were prepared with AI assistance.

D4Vinci: Thanks!

PR #448, opened 2026-09-14
fix(spiders): include params in the request fingerprint

yetval:
Passing query params through `params` gets deduped as if they weren't there. No error, the later pages just never get fetched:

```python
for n in range(3):
    yield Request("https://dummyjson.com/products", params={"limit": 1, "skip": n})
# only product 1 comes back, skip=1 and skip=2 are dropped as duplicates
```

The fingerprint hashes the URL as written, but curl tacks `params` onto it before sending, so all three look identical. `fp_include_kwargs=True` works around it, but pagination through `params` shouldn't need a flag.

Tiny fix so the fingerprint uses the URL curl actually sends. Added a test in `TestRequestProperties` that fails on `dev` today, rest still pass.

Written by me, reviewed by Claude Code.


D4Vinci:
Hi @yetval, nice catch, but why didn't you use `curl_cffi`'s `update_url_params` instead of doing it manually?
It would be cleaner, and manually appending `urlencode(params)` still mishandles fragments, existing query keys, and list/bool values.

Could you use `curl_cffi.requests.utils.update_url_params(self.url, params)` before `canonicalize_url()`? This should make the fingerprint follow the request engine’s URL rules.

Here's an example:
```python
from curl_cffi.requests.utils import update_url_params
from w3lib.url import canonicalize_url

params = self._session_kwargs.get("params")
url = update_url_params(self.url, params) if params else self.url

data = {
    "sid": self.sid,
    "body": body.hex(),
    "method": self._session_kwargs.get("method", "GET"),
    "url": canonicalize_url(url, keep_fragments=keep_fragments),
}
```

Please add tests for these cases and confirm that the helper is available in our minimum supported curl_cffi version.

yetval:
@D4Vinci I completely forgot about that. I have gone ahead and fixed it in the latest commit. Please let me know if there are any other issues you spot!

D4Vinci:
Thanks, @yetval, for your continued contributions to the project!

PR #450, opened 2026-09-16
fix(spiders): skip non-element nodes in sitemap XML instead of crashing on comments

sup3dev:
## What's wrong

`SitemapSpider._get_type` calls `etree.QName(el.tag)` on every node it iterates
(`scrapling/spiders/templates/sitemap.py:105` on `dev`). lxml keeps XML comments
and processing instructions in the parsed tree, and their `.tag` is not a string
(it is `etree.Comment` / `etree.ProcessingInstruction`), so `etree.QName` raises
`ValueError` for them. Nothing catches it, and `_sm_body` iterates `root`
directly — a single comment anywhere under `<urlset>` or `<sitemapindex>`
(e.g. a `<!-- generated by ... -->` note, which real-world sitemaps regularly
carry) kills the whole `_parse_sitemap` callback, and the crawl loses every URL
from that sitemap.

`XMLFeedSpider._iter_nodes` already guards exactly this case with
`isinstance(el.tag, str)` — the sitemap template was missing the same guard.

## What changed

- `_get_type` returns `""` for non-element nodes (the same idiom as the feed
  template). All callers compare the result against known element names
  ("url", "loc", "sitemap"), so comments/PIs are now simply skipped.
- Regression test `test_comments_in_sitemap_are_skipped`: a sitemap with
  comments both under the root and inside `<url>`. Crashes with `ValueError`
  before the fix, passes after.

## Verification

Written by reading the code; the test suite was not run locally — CI (tox)
runs it on this PR.

## AI disclosure (per AI_POLICY.md)

This change was prepared with AI assistance: an agent located the bug by reading
the code and authored the fix and the regression test, which were then reviewed
by the contributor. Degree of assistance: full draft, human review before
submitting.


D4Vinci:
Thanks @sup3dev for catching this one!
It will be released with the next version.

PR #452, opened 2026-09-18
fix(spiders): return no links when a LinkExtractor restriction matches nothing

SulimanAbdulrazzaq:
## Proposed change

`LinkExtractor.extract()` reads the whole page whenever it ends up without scopes:

```python
# scrapling/spiders/links.py, LinkExtractor.extract()
scopes: List[Any] = []
if self.restrict_xpath:
    for xp in self.restrict_xpath:
        scopes.extend(response.xpath(xp))
if self.restrict_css:
    for cs in self.restrict_css:
        scopes.extend(response.css(cs))
if not scopes:
    scopes = [response]
```

`scopes` is also empty when `restrict_css` or `restrict_xpath` is set but matches nothing on the current page. The extractor then returns every link on the page instead of none, so the restriction stops applying on exactly the pages that lack the region.

The usual way to hit this is a `CrawlSpider` rule scoped to a pagination block, which sooner or later reaches the last page:

```python
class Shop(CrawlSpider):
    name = "shop"
    start_urls = ["https://shop.example/products?page=5"]

    def rules(self):
        return [CrawlRule(LinkExtractor(restrict_css=".pagination"))]
```

`Shop().parse()` on a last page that has nav, product and footer links but no `.pagination` block:

```
dev (ce29406): ['https://shop.example/', 'https://shop.example/account', 'https://shop.example/products/41', 'https://shop.example/products/42', 'https://twitter.com/shop', 'https://shop.example/terms']
this PR:       []
```

From there the crawl follows navigation, footer and off-site links that the rule was written to exclude. Scrapy's `LinkExtractor` returns nothing in this case: on the same HTML with Scrapy 2.19.0, `restrict_css=".pagination"` and `restrict_xpaths='//div[@class="pagination"]'` both give `[]`.

### The change

Fall back to the whole page only when no restriction is configured:

```python
if not self.restrict_xpath and not self.restrict_css:
    scopes = [response]
```

An extractor without restrictions still reads the whole page, and a restriction that matches something works as before.

### Tests

`TestRestrict.test_restrict_matching_nothing_extracts_nothing` in `tests/spiders/test_links.py` checks a CSS restriction, an XPath restriction, and both together, each matching nothing. On `dev` it fails because the extractor returns `/nav-link` and `/main-link`; with this change it passes.

Local runs on Linux:

- `pytest tests/spiders/test_links.py tests/spiders/test_templates.py`: 1 failed, 44 passed on `dev` (the new test); 45 passed with this change.
- `pytest tests/spiders`: 460 passed on Python 3.13 and on Python 3.10 (459 on `dev`).
- The two non-browser `tox.ini` runs on Python 3.13 (`-m "asyncio"` and `-m "not asyncio"`, both with `-k "not (DynamicFetcher or StealthyFetcher)"`): the only failures are 52 session tests that launch a browser, which fail with `Executable doesn't exist` because Playwright's Chromium isn't installed in the test environment. `dev` fails the identical 52.
- `ruff check`, `ruff format --check`, `bandit -r -c .bandit.yml`, `vermin -t=3.10- --violations --eval-annotations --no-tips`, `mypy` and `pyright` on `scrapling/`, plus `pre-commit run` on the changed files: all clean.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

No docs change is needed: the `LinkExtractor` reference in `docs/spiders/generic-templates.md` already describes `restrict_css` and `restrict_xpath` as scoping extraction to a region.

### Checklist:

On the unticked boxes: the PR was prepared with AI assistance (see below), though nothing in it is copied from elsewhere. The library change adds no functions; the one new function is a test method, written like the other tests in `test_links.py` without annotations or a docstring.

### AI assistance disclosure

Per [AI_POLICY.md](https://github.com/D4Vinci/Scrapling/blob/main/AI_POLICY.md): this PR was written primarily by Claude Code. It found the defect by reading `LinkExtractor.extract()`, wrote the fix, the regression test and this description, compared Scrapy's behaviour on the same HTML, and ran the tests and checks listed above. A separate Claude Code pass then reviewed the diff and re-ran the red/green test, `tests/spiders`, the reproduction and the Scrapy comparison on the final commit. Replies on this PR will also be written with Claude Code.


D4Vinci:
Nice catch, @SulimanAbdulrazzaq, thanks for finding and solving it cleanly like that!

PR #453, opened 2026-09-18
fix(spiders): ignore a leading byte-order mark in robots.txt and CSV feeds

SulimanAbdulrazzaq:
## Proposed change

The spiders package turns robots.txt and CSV bodies into text with `body.decode(encoding, errors="replace")`. A UTF-8 byte-order mark (`EF BB BF`) survives that as `"﻿"` at the start of the text, and both consumers then misread the first line.

**robots.txt** (`RobotsTxtManager` and `SitemapSpider._robots_body`). protego doesn't strip the BOM, so the first line reads as `"﻿User-agent"`, which isn't a directive it knows, and the whole first group is dropped. For the common file that starts with `User-agent: *`, that group holds every rule. With `robots_txt_obey = True`, `can_fetch()` then allows paths the site disallowed, and its `Crawl-delay`/`Request-rate` are never applied. When the file starts with a `Sitemap:` line instead, `SitemapSpider` loses that sitemap.

```python
>>> from protego import Protego
>>> robots = "﻿User-agent: *\nDisallow: /private\n"
>>> Protego.parse(robots).can_fetch("https://example.com/private", "*")
True
>>> Protego.parse(robots[1:]).can_fetch("https://example.com/private", "*")
False
```

This happens on real sites. Scrapy hit the same protego behaviour in scrapy/scrapy#6195 (labelled `bug`) and fixed it in scrapy/scrapy#7095 by decoding robots.txt with `utf-8-sig`, which drops the BOM before protego sees it. Technical SEO audits keep finding BOM-prefixed robots.txt files in production ([write-up](https://www.gsqi.com/marketing-blog/utf-8-bom-robots-txt/)). Google's [robots.txt documentation](https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt) ("File format" section) says Google ignores a BOM at the start of the file, so these rules apply to Google's crawler while Scrapling skips them.

**CSV feeds** (`CSVFeedSpider`). The BOM becomes part of the first column name, so the value lands under `"﻿title"` and `row["title"]` raises `KeyError`. Excel's "CSV UTF-8" export writes this BOM.

End to end, with bodies that start with a BOM:

```
                                             dev (ce29406)                      this PR
RobotsTxtManager.can_fetch("/admin/secret")  True                               False
RobotsTxtManager crawl-delay                 None                               5.0
SitemapSpider sitemaps from robots.txt       []                                 ['https://example.com/sitemap.xml']
CSVFeedSpider rows                           [{'﻿title': 'First', ...}]    [{'title': 'First', ...}]
```

### The change

Strip one leading BOM before the text is used:

- `RobotsTxtManager._get_parser` and `SitemapSpider._robots_body`: at the `Protego.parse(...)` call.
- `CSVFeedSpider.parse`: right after decoding.

Scrapy's fix changed its decoding to `utf-8-sig`; here the body is decoded with the response's own encoding (the declared charset, or `utf-8`), so the BOM is stripped after decoding instead, which works for any of them. The XML paths (sitemap XML and `XMLFeedSpider`) don't need it: lxml gets the raw bytes and already skips the BOM, which was checked for both. `Selector.json()` is left alone on purpose, since orjson rejects a BOM by design and RFC 8259 tells senders not to add one.

The robots.txt change sits on the `Protego.parse` line, so it doesn't overlap #443, which edits the fetch branch above it; the two merge cleanly. It is also independent of #452 (different files).

### Tests

- `test_robotstxt.py::TestEncoding::test_leading_byte_order_mark_is_ignored`: a BOM before the existing `ROBOTS_BASIC` fixture; `/admin/` must stay disallowed, `/products` allowed, and `Crawl-delay` must still read `2.0`.
- `test_sitemap.py::TestRobotsTxt::test_parse_sitemap_robots_with_leading_byte_order_mark`: a robots.txt that starts with a BOM and a `Sitemap:` line yields that sitemap.
- `test_feed.py::TestCSVFeedSpider::test_leading_byte_order_mark_stays_out_of_the_first_column`: a BOM before `CSV_BODY` gives the same first row as without it.

All three fail on `dev` and pass with this change.

Local runs on Linux:

- `pytest tests/spiders/test_robotstxt.py tests/spiders/test_sitemap.py tests/spiders/test_feed.py`: 3 failed, 68 passed on `dev` (the three new tests); 71 passed with this change.
- `pytest tests/spiders`: 462 passed on Python 3.13 and on Python 3.10 (459 on `dev`).
- The two non-browser `tox.ini` runs on Python 3.13 (`-m "asyncio"` and `-m "not asyncio"`, both with `-k "not (DynamicFetcher or StealthyFetcher)"`): the only failures are 52 session tests that launch a browser, which fail with `Executable doesn't exist` because Playwright's Chromium isn't installed in the test environment. `dev` fails the identical 52.
- `ruff check`, `ruff format --check`, `bandit -r -c .bandit.yml`, `vermin -t=3.10- --violations --eval-annotations --no-tips`, `mypy` and `pyright` on `scrapling/`, plus `pre-commit run` on the changed files: all clean.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

### Checklist:

On the unticked boxes: the PR was prepared with AI assistance (see below), though nothing in it is copied from elsewhere. The library change adds no functions; the new functions are test methods, written like most tests in these files without annotations or docstrings.

### AI assistance disclosure

Per [AI_POLICY.md](https://github.com/D4Vinci/Scrapling/blob/main/AI_POLICY.md): this PR was written primarily by Claude Code. It found the defect by tracing how the spiders decode response bodies, confirmed protego's behaviour, wrote the fix, the regression tests and this description, and ran the tests and checks listed above. A separate Claude Code pass then reviewed the diff, checked the Scrapy, Google and audit references, and re-ran the red/green tests, `tests/spiders` and the end-to-end reproduction on the final commit. Replies on this PR will also be written with Claude Code.


D4Vinci:
Thanks, @SulimanAbdulrazzaq, for catching and solving this!

Issue #447, opened 2026-09-14
Same-origin iframe documents are silently dropped from pierce_shadow snapshots — fix ready, PR welcome?


ZeroCode00:
First — the new shadow-piercing snapshot on `dev` is really nicely done; the native-serialization approach is much cleaner than what I attempted in #441.

**What happened**: with `pierce_shadow=True`, same-origin iframe documents are not traversed, so their content is silently missing from the snapshot. I saw the note in the docs that this is intentional — but I wanted to flag it with concrete evidence, because in the wild this pattern is more common than it looks:

- **TinyMCE's official demo** renders its entire editing surface in a same-origin `srcdoc` iframe (with a shadow root inside). On current `dev` with `pierce_shadow=True`, **none of the ~1,750 characters of editor content is selectable** — even though the `srcdoc` attribute text remains, which can mislead users into thinking the data is there. CKEditor 4's demo follows the same pattern.
- On my synthetic nested-shadow page (13 extraction targets: light DOM, shadow depth 2–4, slots, shadow-inside-iframe, closed root), current `dev` reaches 11/13; the 2 misses are the iframe case and the closed root.

**Environment**: Python 3.13.15 (64-bit venv) on Windows 11 Pro 25H2, scrapling 0.4.15 from `dev` (`6e64404`), Playwright 1.62.0 with bundled Chromium and system Chrome (both reproduced).

The good news: same-origin frames don't require any architectural change — `contentDocument` is reachable from the main frame's context, so recursion stays inside the existing single-evaluate walker. I have a working implementation on top of your `dev`:

- Frame documents are serialized into `<iframe-document>` siblings right after their `<iframe>` (iframes are raw-text in HTML parsing, so frame content can't be nested inside them);
- Recursion covers frames-in-frames and shadow-inside-frame; cross-origin frames stay a documented boundary, skipped without errors;
- Walker type checks are now realm-safe (`instanceof` fails across frame realms);
- **55/55 tests pass** (4 new + all existing shadow tests), validated live against the TinyMCE demo (patch: 1,754 chars reachable; without: 0).

Branch: `ZeroCode00/Scrapling@feat/same-origin-iframe-snapshot`, evidence pack with full test logs: branch `docs/iframe-pr-notes`.

**Would you like me to send this as a PR against `dev`?** Happy to adjust naming, placement, or scope however you prefer.

(AI disclosure: the implementation and tests were AI-assisted, under my direction and review; all numbers are from my local runs.)

[...]

PR #454, opened 2026-09-19
fix(browsers): preserve same-origin iframe documents in shadow snapshots

Karthik9849309055:
## What

When `pierce_shadow=True`, same-origin iframe documents were silently skipped, making content (e.g. TinyMCE, CKEditor editors) unreachable even though the `srcdoc` attribute made it appear present.

## How

- After copying each `<iframe>`, the walker now accesses `contentDocument` and appends its content as an `<iframe-document>` sibling element.
- Cross-origin frames are silently skipped via `try/catch` (documented boundary, no change there).
- Replaced `instanceof HTMLSlotElement/HTMLTemplateElement` with `node.localName` checks — `instanceof` fails across frame realms.

## Tests

4 new tests in `tests/fetchers/test_shadow_iframe.py`:
- Basic iframe content preserved
- Shadow-inside-iframe pierced correctly
- Nested iframe recursion works
- Cross-origin iframe skipped without error

> AI disclosure: implementation was AI-assisted, under my direction and review.


D4Vinci:
Thanks for the contribution. I’m closing this PR because combining iframe documents with the main response needs a different design.

Testing confirmed that relative links inside a frame resolve against the parent page’s URL, and hidden iframe content can survive Markdown cleaning. Frame traversal also misses some cases, and the new element checks can break existing shadow extraction.

I don’t want to expand `pierce_shadow` to automatically include separate documents. I’ll keep #447 open to consider explicit frame capture that preserves each document’s URL and identity.

PR #458, opened 2026-09-23
fix(spiders): route sitemap URLs to parse() instead of the sitemap parser

SulimanAbdulrazzaq:
## Proposed change

With no `rules()`, or with a matching `CrawlRule` that has no callback, `SitemapSpider` never calls `parse()` for the URLs listed in a sitemap.

`start_requests()` creates each sitemap request with `callback=self._parse_sitemap`. `_dispatch()` then followed each listed URL with `response.follow(url)` or `response.follow(url, callback=rule.callback)`. When no callback is given, `Response.follow()` falls back to `self.request.callback`, so every page request inherited `_parse_sitemap`. The HTML pages were then parsed as sitemaps, which logs `Unknown sitemap root element: 'html'` and yields nothing.

That contradicts the docs (`docs/spiders/generic-templates.md`: "If `rules()` returns an empty list, every URL is routed to the spider's `parse()` method") and `CrawlRule`'s docstring ("Falls back to the spider's default ``parse()``").

The fix is limited to the template. `_dispatch()` is now a regular method and passes `self.parse` explicitly when there are no rules, or when the matching rule has no callback. `Response.follow()` is unchanged, since its fallback to the parent request's callback is general behaviour. `_dispatch()` has a single caller (`_parse_sitemap`), which already calls it through `self`.

The existing `test_no_rules_means_all_urls_fall_through` builds its response with `Request(url, sid="default")`, a request with no callback, which a real crawl never produces. That is why it didn't catch this. I left the helper as it is. That test now asserts `callback == spider.parse` instead of `None`, since the callback is explicit now. Two new tests set the response's request to `callback=spider._parse_sitemap`, as in a real crawl: one for the no-rules case, and one for a mix of a rule with a callback and a rule without one.

End to end, on Windows with Python 3.12: a `SitemapSpider` without rules crawled a local `http.server` serving a sitemap with two HTML pages.
- Before: `items scraped: 0`, two `Unknown sitemap root element: 'html'` warnings, `stats: requests=3 items=0`.
- After: both pages reach `parse()`, `items scraped: 2`, `stats: requests=3 items=2`.

This bug was found by reading the code. There is no issue for it.

**AI assistance:** This PR was written by an AI agent (Claude Code, Claude Opus models): investigation, fix and tests; reviewed by the coordinating agent before submission.

### Type of change:

- [ ] Dependency upgrade
- [ ] Documentation change?

### Additional information

- This PR fixes or closes an issue: fixes # (none, self-found)

### Checklist:
* [ ] This pull request is all my own work -- I have not plagiarized. (Written with AI assistance, as disclosed above; nothing is copied from other sources.)


SulimanAbdulrazzaq:
@D4Vinci when you have a moment, could you take a look at this one? With no `rules()`, or with a matching `CrawlRule` that has no callback, `SitemapSpider` sends the URLs listed in a sitemap back to the sitemap parser, so `parse()` is never called for them. The branch merges cleanly and CI is green.

D4Vinci:
Good catch @SulimanAbdulrazzaq !
Thanks a lot for finding and solving this one. It will be merged and added to the next major release.

